PR #125, opened 2026-04-07
feat(subsonic): per-server music folder filter and sidebar picker

cucadmuh:
# Feature: Subsonic music library (folder) scope

## Summary

This PR adds **per-server browsing scope** for Subsonic-compatible servers that expose **multiple music folders** (`getMusicFolders`). The user picks either **all libraries** (default, no `musicFolderId` on the wire) or **exactly one** folder. The choice is **remembered per server**; catalog APIs append **`musicFolderId`** where the Subsonic spec supports it, and library-oriented screens **refetch** when the scope changes.

The sidebar shows a **compact dropdown** (portal + fixed positioning so it is not clipped by the nav scroller) when the server reports **more than one** folder. **Single-select** behaviour with a checkmark on the active row; **>10 folders** caps panel height so the list scrolls.

### Symptom

1. Some backends (e.g. **Navidrome** with multiple libraries, or classic Subsonic **music folders**) expose **several roots** under one account.
2. Without a scope control, the app always behaves like **“all folders merged”**, which is noisy on large multi-library setups and does not match how users often think about **separate libraries** (e.g. main vs explicit content).

### Why not multi-select

Subsonic filtering is **one optional `musicFolderId` per request**, not a list. Supporting “several checked folders” would imply **client-side merging** of multiple API calls, pagination quirks, and inconsistent counts. This PR intentionally supports **all** or **one** only.

### 1. State: `authStore`

- **`musicFolders`**: list for the **active** server, fetched after login / server switch (**not** persisted).
- **`musicLibraryFilterByServer`**: map `serverId → 'all' | folderId` (**persisted** via existing zustand `partialize`; volatile fields stripped).
- **`musicLibraryFilterVersion`**: incremented in **`setMusicLibraryFilter`** so pages can depend on a simple primitive and reload catalog data (**not** persisted).
- **`setMusicFolders`**: if the stored folder id **no longer exists** after a refetch (server config changed), scope falls back to **`all`** for that server.

### 2. API: `libraryFilterParams()` and `getMusicFolders`

- **`getMusicFolders()`** parses `getMusicFolders.view` into `{ id, name }[]`.
- **`libraryFilterParams()`** returns `{}` or `{ musicFolderId }` from the active server’s entry in **`musicLibraryFilterByServer`**.

Merged into requests for:

- **`getAlbumList2.view`** — album lists (incl. random, newest, by genre, etc.)
- **`getRandomSongs.view`**
- **`getArtists.view`**
- **`getStarred2.view`**
- **`search3.view`**

Other calls (e.g. **`getGenres.view`**, **`getAlbum.view` by id**, **`getArtist.view` by id**) stay **unscoped** by design: genre lists and direct entity fetches are either global on the server or already anchored by id.

### 3. Lifecycle

- **`App.tsx`**: after login / **`activeServerId`** change, load folders and **`setMusicFolders`**; errors clear the list.
- **Screens and search surfaces** that show **library-derived** data subscribe to **`musicLibraryFilterVersion`** (and keep their other deps) so a scope change triggers a **refetch** or equivalent refresh.

### 4. Sidebar UX

- Dropdown rendered with **`createPortal(..., document.body)`** and width aligned to the trigger **`getBoundingClientRect()`** so the panel matches the button and avoids **sidebar `overflow` clipping**.
- **“All libraries”** mode uses a **plain** trigger (label only + chevron, no boxed chrome).
- **Single folder** selected: **framed** trigger, **note** icon, **folder name** under the **Library** caption.
- **More than 10** folders: panel gets a **max-height** class so roughly **one “All” row + ten folder rows** show and the rest **scroll** (capped by **`70vh`**).

### 5. i18n

Strings for scope UI added for **en, ru, de, fr, nb, nl, zh** to match existing locale coverage.

### 6. Tooltips (follow-up)

Opening the library dropdown from a **`data-tooltip`** control left the hover tooltip visible over the portal menu. **`TooltipPortal`** now clears the overlay on **`mousedown`** (capture) for any anchor with **`data-tooltip`**. The library trigger also omits **`data-tooltip`** while the menu is **`aria-expanded`**.

## Files changed

**Store / API / shell**

- `src/store/authStore.ts` — folders list, per-server filter map, version bump, persist `partialize`, validation when folder list updates.
- `src/api/subsonic.ts` — `getMusicFolders`, `libraryFilterParams`, pass-through on scoped endpoints listed above.
- `src/App.tsx` — fetch music folders on login / server change.

**UI**

- `src/components/Sidebar.tsx` — library scope trigger, portal dropdown, conditional **`data-tooltip`** when closed.
- `src/components/TooltipPortal.tsx` — dismiss tooltip on **`mousedown`** on **`[data-tooltip]`** anchors (avoids overlap with click-open menus).
- `src/styles/layout.css` — trigger, panel, scroll-cap styles.

**Pages** (deps on **`musicLibraryFilterVersion`** and/or scoped data)

- `src/pages/Home.tsx`
- `src/pages/Albums.tsx`
- `src/pages/Artists.tsx`
- `src/pages/ArtistDetail.tsx`
- `src/pages/Favorites.tsx`
- `src/pages/NewReleases.tsx`
- `src/pages/RandomAlbums.tsx`
- `src/pages/RandomMix.tsx`
- `src/pages/Statistics.tsx`
- `src/pages/SearchResults.tsx`
- `src/pages/GenreDetail.tsx`
- `src/pages/Genres.tsx`
- `src/pages/AdvancedSearch.tsx`
- `src/pages/LabelAlbums.tsx`

**Other components**

- `src/components/Hero.tsx`
- `src/components/LiveSearch.tsx`
- `src/components/MobileSearchOverlay.tsx`

**Locales**

- `src/locales/en.ts`
- `src/locales/ru.ts`
- `src/locales/de.ts`
- `src/locales/fr.ts`
- `src/locales/nb.ts`
- `src/locales/nl.ts`
- `src/locales/zh.ts`

**Total: 30 files** (29 for folder scope + **`TooltipPortal.tsx`**; **`Sidebar.tsx`** updated in both commits).

## How to test

1. Use a server with **≥ 2** music folders (Navidrome libraries or Subsonic folders).
2. Log in; **Expected:** sidebar shows **Library** control with **All libraries** + each folder.
3. Pick **one** folder; **Expected:** subtitle shows its name; **albums / artists / favorites / search / starred / random** style views reflect **only** that folder (within API support).
4. Switch back to **all**; **Expected:** plain trigger, merged catalogue again.
5. Change **active server** in multi-server setup; **Expected:** folder list reloads; per-server choice is **restored** when you return.
6. Collapse sidebar; **Expected:** picker hidden (unchanged single-folder servers still show normal section label only when picker not shown).
7. If you can simulate **>10** folders (or mock), **Expected:** dropdown scrolls inside the panel without blowing past the viewport height.
8. Optional: remove a folder on the server that was selected; **Expected:** after refetch, scope resets to **all** if the id is gone.
9. Open the library menu from the sidebar control; **Expected:** no **hover tooltip** stays on screen on top of the dropdown.

## Commits (high level)

1. **`feat(subsonic): per-server music folder filter and sidebar picker`** — store, API params, App fetch, sidebar dropdown, page refresh hooks, i18n, layout CSS.
2. **`fix(ui): dismiss tooltip on mousedown; hide library dropdown hint while open`** — **`TooltipPortal`** + **`Sidebar`** (see §6 above).

## Notes for reviewers

- **Compatibility:** `musicFolderId` is optional in the Subsonic API; servers with a **single** folder behave as before (picker UI is hidden when `musicFolders.length <= 1`).
- **Playback:** stream URLs are **not** altered here; scope affects **browsing and search**, not per-track auth.
- **Scope:** endpoints without `musicFolderId` in this PR are **intentional** where the spec or product does not require folder filtering for that call.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -53,6 +53,7 @@ import AppUpdater from './components/AppUpdater';
 import { version } from '../package.json';
 import { useConnectionStatus } from './hooks/useConnectionStatus';
 import { useAuthStore } from './store/authStore';
+import { getMusicFolders } from './api/subsonic';
 import { useOfflineStore } from './store/offlineStore';
 import { initHotCachePrefetch } from './hotCachePrefetch';
 import { usePlayerStore, initAudioListeners } from './store/playerStore';
@@ -82,8 +83,27 @@ function AppShell() {
   const navigate = useNavigate();
   const location = useLocation();
   const serverId = useAuthStore(s => s.activeServerId ?? '');
+  const isLoggedIn = useAuthStore(s => s.isLoggedIn);
+  const activeServerId = useAuthStore(s => s.activeServerId);
+  const setMusicFolders = useAuthStore(s => s.setMusicFolders);
   const offlineAlbums = useOfflineStore(s => s.albums);
   const hasOfflineContent = Object.values(offlineAlbums).some(a => a.serverId === serverId);
+
+  useEffect(() => {
+    if (!isLoggedIn || !activeServerId) return;
+    let cancelled = false;
+    (async () => {
+      try {
+        const folders = await getMusicFolders();
+        if (!cancelled) setMusicFolders(folders);
+      } catch {
+        if (!cancelled) setMusicFolders([]);
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
+  }, [isLoggedIn, activeServerId, setMusicFolders]);
 
   // Auto-navigate to offline library when no connection but cached content exists
   const prevConnStatus = useRef(connStatus);
--- a/src/api/subsonic.ts
+++ b/src/api/subsonic.ts
@@ -40,6 +40,15 @@ async function api<T>(endpoint: string, extra: Record<string, unknown> = {}, tim
   return data as T;
 }
 
+/** Optional `musicFolderId` when the user narrowed browsing to one Subsonic library (see `getMusicFolders`). */
+export function libraryFilterParams(): Record<string, string | number> {
+  const { activeServerId, musicLibraryFilterByServer } = useAuthStore.getState();
+  if (!activeServerId) return {};
+  const f = musicLibraryFilterByServer[activeServerId];
+  if (f === undefined || f === 'all') return {};
+  return { musicFolderId: f };
+}
+
 // ─── Types ────────────────────────────────────────────────────
 export interface SubsonicAlbum {
   id: string;
@@ -139,6 +148,11 @@ export interface SubsonicGenre {
   albumCount: number;
 }
 
+export interface SubsonicMusicFolder {
+  id: string;
+  name: string;
+}
+
 export interface SubsonicArtistInfo {
   biography?: string;
   musicBrainzId?: string;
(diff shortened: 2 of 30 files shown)


Psychotoxical:
Merged manually into main with review fixes applied (data-tooltip, single set(), libraryScopeHint cleanup, Genres deps).

PR #175, opened 2026-04-13
fix(fullscreen-player): stop mesh blob and portrait animations in no-compositing mode

kilyabin:
In software compositing (WEBKIT_DISABLE_COMPOSITING_MODE=1), WebKit renders entirely on the CPU. The mesh blob animations were not covered by the no-compositing overrides: blob-a occupies 130×130% of the viewport, so each frame the CPU had to composite a surface larger than the screen. On high-DPI displays (e.g. 8K) this drops the fullscreen player to ~10 FPS.

- Stop mesh blob pan animations (static gradients are preserved)
- Stop portrait drift animation (filter was already removed)
- Remove box-shadow from the seekbar played bar (its width changes on every playback tick; in software mode each change triggers a box-shadow repaint)

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -3526,10 +3526,11 @@ html.no-compositing .fs-close {
   background: rgba(0, 0, 0, 0.45);
 }
 
-/* Portrait: remove CSS filter (drop-shadow) and will-change */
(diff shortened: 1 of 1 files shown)


kilyabin:
build not checked, but im think all is ok

PR #211, opened 2026-04-19
feat: floating play bar toggle with liquid glass

kveld9:
# Floating Player Bar with Liquid Glass aesthetic

Adds an optional floating player bar with an authentic liquid glass effect that refracts background content through a translucent, blurred surface.

---

### State Management
- Added `floatingPlayerBar` boolean state and setter to `themeStore.ts`
- State persists via zustand's `persist` middleware

### Settings UI
- Added toggle switch in **Settings → Appearance → Visual Options**
- Positioned after the "Show Bitrate" option

### Localization
Added `floatingPlayerBar` and `floatingPlayerBarSub` keys to all 8 locale files: `de`, `en`, `es`, `fr`, `nb`, `nl`, `ru`, `zh`

### App Shell Integration
- `App.tsx` reads `floatingPlayerBar` state and applies `data-floating-player-bar` attribute to `.app-shell`

### Liquid Glass CSS (`layout.css`)
- Fixed positioning with dynamic margins based on sidebar/queue widths
- `backdrop-filter: blur(28px) saturate(200%) brightness(1.08) contrast(108%)`
- Asymmetric borders — top brighter than bottom to simulate realistic light incidence
- Cool-tint background gradient with subtle blue refractive tone
- 1px specular highlight line at top edge via `::after`
- Glass treatment for child elements:
  - **Play button** — semi-transparent with nested blur
  - **Volume slider** — glass track and thumb with specular highlights
  - **Icons** — 60% opacity transitioning to 100% on hover
  - **Album art** — integrated glass border

### Theme Override (`theme.css`)
- Added conditional glass styling for the play button, scoped to `[data-floating-player-bar]` so it only applies when the feature is active

---

## Behavior

| State | Result |
|-------|--------|
| Toggle **OFF** | Player bar renders with default theme styling |
| Toggle **ON** | Player bar floats above content with liquid glass effect, refracting background colors through the translucent surface |

---

### TypeScript / TSX
- `src/store/themeStore.ts` — Added `floatingPlayerBar` state and setter
- `src/pages/Settings.tsx` — Added toggle in Visual Options section
- `src/App.tsx` — Applied `data-floating-player-bar` attribute to `.app-shell`

### CSS
- `src/styles/layout.css` — Liquid glass styling for floating bar, play button, volume slider, icons, and album art
- `src/styles/theme.css` — Play button glass override (conditional on `data-floating-player-bar`)

### Localization
- `src/locales/de.ts`
- `src/locales/en.ts`
- `src/locales/es.ts`
- `src/locales/fr.ts`
- `src/locales/nb.ts`
- `src/locales/nl.ts`
- `src/locales/ru.ts`
- `src/locales/zh.ts`

### Screenshoots

{098F4303-C532-4377-8638-4FD69E6E7E30}

{BF3E49C4-4108-413B-B7FB-06E003B58BB1}

{1C83899D-ECF6-4CB4-BE72-BCB42C87F637}

{068C6D91-5656-4A20-9FDD-8AC29DCD5785}

{DB12C80D-CC96-4283-8A89-A21D8DF28263}

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "psysonic",
-  "version": "1.34.13",
+  "version": "1.41.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "psysonic",
-      "version": "1.34.13",
+      "version": "1.41.0",
       "dependencies": {
         "@fontsource-variable/dm-sans": "^5.2.8",
         "@fontsource-variable/figtree": "^5.2.10",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -143,6 +143,7 @@ function AppShell() {
   const useCustomTitlebar = useAuthStore(s => s.useCustomTitlebar);
   const linuxWebkitKineticScroll = useAuthStore(s => s.linuxWebkitKineticScroll);
   const setEntityRatingSupport = useAuthStore(s => s.setEntityRatingSupport);
+  const floatingPlayerBar = useThemeStore(s => s.floatingPlayerBar);
   const offlineAlbums = useOfflineStore(s => s.albums);
   const hasOfflineContent = Object.values(offlineAlbums).some(a => a.serverId === serverId);
 
@@ -350,6 +351,7 @@ function AppShell() {
       data-mobile-player={isMobilePlayer || undefined}
       data-titlebar={(IS_LINUX && useCustomTitlebar && !isWindowFullscreen && !isTilingWm) || undefined}
       data-fullscreen={isWindowFullscreen || undefined}
+      data-floating-player-bar={floatingPlayerBar || undefined}
       style={{
         '--sidebar-width': isMobile ? '0px' : (isSidebarCollapsed ? '72px' : 'clamp(200px, 15vw, 220px)'),
         '--queue-width': isMobile ? '0px' : (isQueueVisible ? `${queueWidth}px` : '0px')
--- a/src/locales/de.ts
+++ b/src/locales/de.ts
@@ -731,6 +731,8 @@ export const deTranslation = {
     playlistCoverPhotoSub: 'Zeigt Coverfoto-Raster in der Playlist-Detailansicht',
     showBitrate: 'Bitrate anzeigen',
     showBitrateSub: 'Audio-Bitrate in Track-Listen anzeigen',
+    floatingPlayerBar: 'Schwebende Wiedergabeleiste',
+    floatingPlayerBarSub: 'Aktiviert schwebende Wiedergabeleiste mit Liquid-Glass-Stil und Brechungseffekt',
     uiScaleTitle: 'Interface-Skalierung',
     uiScaleLabel: 'Zoom',
   },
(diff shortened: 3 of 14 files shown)


Psychotoxical:
Hey @kveld9, thanks a lot for putting this together — the screenshots look really nice and the idea of an opt-in liquid-glass player bar is genuinely tempting. I'd love to land it in some form, but I have a few concerns about the current implementation that I think are worth working through before merge.

### Performance on WebKitGTK (Linux)

The biggest one. Linux runs in WebKitGTK with `WEBKIT_DISABLE_COMPOSITING_MODE=1` (set in `tauri:dev` and effectively the default fallback in production), which means every compositing layer is composed in software. We've already had to strip `filter: blur()` out of the hero background, the album-detail background and the fullscreen-player mesh blobs because they tanked frame times — see commit history for the v1.34.4 CPU optimisations.

This PR re-introduces the same cost on a much more sensitive surface:
- `backdrop-filter: blur(28px) saturate(200%) brightness(1.08) contrast(108%)` on a permanently-visible bar
- Plus a second `backdrop-filter: blur(12px) saturate(180%)` on the play button
- Plus a multi-layer `box-shadow` stack (5 layers, including `inset 0 0 40px`)
- Plus `transition: all 0.4s` (which fires on any property change, multiplying repaints)

That stack will most likely cause the player bar to drop frames during playback on every Linux machine that isn't a Wayland Mutter host with full hardware acceleration. Would it be acceptable to dial these down significantly — say `blur(8-10px)` only, drop the saturate/brightness/contrast filters, and replace `transition: all` with a targeted `transition: background, box-shadow`?

### Theme integration

A few things break the theming system:
1. **Hardcoded `rgba(255, 255, 255, …)` everywhere.** Light themes (Latte, Nord Snowstorm) have a light player background — the white-on-white glass overlay washes out completely there. Could you sample the surface colors via `var(--bg-player)`, `var(--text-primary)`, etc., or via `color-mix(in srgb, var(--bg-player) 80%, white 20%)` so it adapts?
2. **Duplicated theme selectors.** `[data-theme] .app-shell[data-floating-player-bar] …` already matches every theme — the explicit `mocha`/`latte`/`frappe`/`macchiato` lines after it are redundant.
3. **`!important` everywhere.** Once the rule lives under `[data-floating-player-bar]` it's already specific enough; `!important` makes future overrides painful and can be dropped.
4. **Volume slider doesn't use `--volume-accent`** — that variable is intentional (volume color is independent of the theme accent).

### Layout

- `border-radius: 100px` on the container looks great in the screenshots, but the internal album art and buttons sit inside it without any matching clipping — at very narrow widths the corners may show square content under a rounded mask.
- Switching the bar from grid-row to `position: fixed` should also be tested with the custom titlebar (`data-titlebar`), the mobile player layout, and the Linux fullscreen WM mode, since those rely on the grid template.

### Nit

`package-lock.json` got bumped to `1.41.0` — that part is actually correct (main's lockfile is just stale), no need to revert. Just flagging it isn't a "your PR" change.

---

Sorry for the wall of text — I really do want a version of this in. Happy to pair on a slimmer, performance-friendly variant if you'd like, or to take the existing branch and rework it together. Either way, thanks for the contribution!

kilyabin:
It looks pretty good, but I agree with @Psychotoxic—this could cause performance issues on Linux. Personally, I think it looks a bit out of place with the main design and doesn’t quite fit in. Maybe we should try making the other elements in the same style with rounded corners.

Psychotoxical:
I'm checking now how to implement these version exclusively for mac and windows? OK

Psychotoxical:
image

Works. Linux solid bar, win and mac glassy bar.

Thanks for your hard work on this one @kveld9  !!!

PR #216, opened 2026-04-19
feat: floating player bar with toggle

kveld9:
## Summary

Improvements to the floating player bar: better theming, proper album art alignment.

## What changed

- **Floating bar styling** – solid background that adapts to light/dark themes, accent-colored border, and elevation shadow
- **Album art** – rounded corners now match the bar’s shape on the left side
- **Volume section** – buttons and slider are vertically centered

## Result

The floating player bar now looks consistent across all themes and separates properly from background content.

## Notes
- I decided against using the blur effect due to performance issues in WebKit2GTK. I think this implementation is a solid option right now, although there’s still room for improvement.

## Screenshoots

image

image

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,6 @@
         "@tauri-apps/plugin-updater": "^2.10.0",
         "@tauri-apps/plugin-window-state": "^2.4.1",
         "axios": "^1.7.7",
-        "colorthief": "^3.3.1",
         "i18next": "^25.8.16",
         "lucide-react": "^0.462.0",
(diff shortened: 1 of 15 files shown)


kilyabin:
I think we could create two versions: one for Linux with a solid background, and one for Windows/macOS with a "liquid glass" effect.

Psychotoxical:
> I think we could create two versions: one for Linux with a solid background, and one for Windows/macOS with a "liquid glass" effect.

Thats possible. Checking this one first.

PR #217, opened 2026-04-19
feat(linux): Auto-detect GPU vendor and configure DMA-BUF renderer

kveld9:
### Summary

Implements automatic GPU vendor detection on Linux to optimize WebKitGTK performance.
Configures `WEBKIT_DISABLE_DMABUF_RENDERER` based on detected GPU:

- **NVIDIA**: Disables DMA-BUF (proprietary drivers have issues with WebKitGTK)
- **Intel/AMD**: Enables DMA-BUF (Mesa drivers handle it well)
- **Unknown**: Defaults to safe mode (disabled)

---

#### `src-tauri/src/main.rs`

**Added `GpuVendor` enum** for type-safe GPU vendor representation:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
enum GpuVendor { Nvidia, Intel, Amd }
```

**Added `detect_gpu_vendor()` function** with multi-layer detection strategy:

| Priority | Method                            | Notes                                               |
|----------|-----------------------------------|-----------------------------------------------------|
| 1        | `/proc/driver/nvidia/version`     | Most reliable for NVIDIA proprietary drivers        |
| 2        | `/sys/class/drm/card*/device/vendor` | PCI vendor IDs (0x10de, 0x8086, 0x1002/0x1022)  |
| 3        | `lspci -nn`                       | Requires pciutils; only matches VGA/Display/3D      |
| 4        | `glxinfo -B`                      | Last resort; requires active X11/display            |

**Modified `main()`** to configure DMA-BUF based on detected GPU:

```rust
match detect_gpu_vendor() {
    Some(GpuVendor::Nvidia)                    => set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1"),
    Some(GpuVendor::Intel) | Some(GpuVendor::Amd) => set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "0"),
    None                                       => set_var("WEBKIT_DISABLE_DMABUF_RENDERER", "1"), // safe default
}
```

---

### Why This Matters

WebKitGTK's DMA-BUF renderer behavior differs significantly between GPU vendors:

- **NVIDIA proprietary drivers**: DMA-BUF causes visual glitches, lag, and frame drops.
  Disabling it forces the CPU-copy path, which is more stable.
- **Intel/AMD (Mesa)**: DMA-BUF enables zero-copy frame passing from GPU to WebView,
  improving performance and reducing CPU usage.

Previously, the app always used safe defaults or required manual user configuration.
Now it automatically optimizes for the detected hardware while still allowing user override
via environment variables.

---

### Technical Notes

- **Detection order**: Prioritized from most reliable (NVIDIA kernel module) to least (glxinfo requiring active display)
- **lspci filtering**: Only matches display controllers (`vga`, `display`, `3d controller`) to avoid false positives from AMD audio devices
- **Safety**: `set_var` is called before the Tauri runtime spawns threads, so there are no concurrent access issues
- **User override**: All variables are only set if not already defined, preserving user control

---

### Testing

- [x] Compiles with `cargo check`
- [x] Enum-based matching is type-safe and exhaustive
- [x] Each detection layer handles missing tools gracefully

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -1,16 +1,122 @@
 // Prevents additional console window on Windows in release, DO NOT REMOVE!!
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
+#[cfg(target_os = "linux")]
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum GpuVendor {
+    Nvidia,
+    Intel,
+    Amd,
+}
+
+#[cfg(target_os = "linux")]
+fn detect_gpu_vendor() -> Option<GpuVendor> {
+    use std::fs;
+    use std::process::Command;
+
+    // Check for NVIDIA driver presence (most reliable for proprietary drivers)
+    if fs::metadata("/proc/driver/nvidia/version").is_ok() {
+        return Some(GpuVendor::Nvidia);
+    }
+
+    // Check sysfs DRM vendor IDs (works for most integrated/discrete GPUs)
+    let vendor_paths = [
+        "/sys/class/drm/card0/device/vendor",
+        "/sys/class/drm/card1/device/vendor",
+    ];
+
+    for path in &vendor_paths {
+        if let Ok(vendor_id) = fs::read_to_string(path) {
+            let vendor_id = vendor_id.trim();
+            // PCI vendor IDs: NVIDIA=0x10de, Intel=0x8086, AMD=0x1002/0x1022
+            match vendor_id {
+                "0x10de" => return Some(GpuVendor::Nvidia),
+                "0x8086" => return Some(GpuVendor::Intel),
+                "0x1002" | "0x1022" => return Some(GpuVendor::Amd),
+                _ => {}
+            }
+        }
+    }
+
+    // Fallback: try lspci to detect GPU vendor (requires pciutils)
+    if let Ok(output) = Command::new("lspci").args(["-nn"]).output() {
+        if output.status.success() {
+            let stdout = String::from_utf8_lossy(&output.stdout);
+            for line in stdout.lines() {
+                let line_lower = line.to_lowercase();
+                // Only match VGA/Display controllers, not audio or other devices
+                let is_display = line_lower.contains("vga")
+                    || line_lower.contains("display")
+                    || line_lower.contains("3d controller");
+
+                if is_display && line_lower.contains("nvidia") {
+                    return Some(GpuVendor::Nvidia);
+                }
+                if is_display
+                    && (line_lower.contains("intel") || line_lower.contains("i915"))
+                {
+                    return Some(GpuVendor::Intel);
+                }
+                if is_display && (line_lower.contains("amd") || line_lower.contains("radeon"))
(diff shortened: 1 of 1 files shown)


Psychotoxical:
Thanks for the contribution, @kveld9 — merged!

Before pushing the green button I spent close to an hour on a **blind A/B test** on my own Linux + NVIDIA proprietary box (the exact config this PR targets). 4 runs, randomised order I didn't know at the time — 2× with the DMA-BUF renderer disabled (this PR's behaviour on NVIDIA) and 2× with it enabled. Same usage parcour each run: album browsing, playback, skip-burst, seekbar clicks, fullscreen player with mesh blobs, queue scroll/drag, lyrics pane.

| Run | Variant | CPU playback | CPU fullscreen | CPU browsing | Subjective fluffiness | Glitches |
|-----|---------|--------------|----------------|--------------|-----------------------|----------|
| 1 | DMA-BUF ON   | 20 → 8%   | 18 → 10%   | 7–17% | 7/10 | none |
| 2 | DMA-BUF OFF  | 17 → 6.5% | —          | 10–15% | 8/10 | none |
| 3 | DMA-BUF OFF  | 18 → 8%   | 18 → <10%  | 18 → 9% | 8/10 | none |
| 4 | DMA-BUF ON   | 19 → 10%  | 19 → <10%  | 16 → 9% | 7/10 | none |

Clean 2:2 split on the blind fluffiness scores — both "OFF" runs landed at 8, both "ON" runs at 7, without me knowing which variant was running. CPU numbers line up with the subjective impression: the lowest playback-state CPU and the lowest browsing peak both fell on the "OFF" runs. No glitches on either variant, so the toggle isn't harmful when DMA-BUF stays on either — it's a small but genuine quality-of-life win for the NVIDIA + proprietary crowd.

I pushed a small follow-up commit on `main` (36caf3a1) with a few tweaks that came up during the review:

- Dropped the `lspci` / `glxinfo` subprocess fallbacks. The `/proc/driver/nvidia/version` + sysfs scan covered every realistic case in my testing, and spawning processes in `main()` added startup latency for no measurable benefit (`glxinfo` also isn't always installed — `mesa-utils` isn't a hard dep).
- Removed `0x1022` from the AMD check — that's AMD's host/chipset vendor ID, not GPU. Only `0x1002` (ATI/Radeon graphics) belongs there.
- Unknown vendor → leave `WEBKIT_DISABLE_DMABUF_RENDERER` unset instead of forcing `=1`. VMs, ARM SBCs and anything exotic keep the WebKitGTK default and don't get regressed by a guess.
- Replaced the hardcoded `card0` / `card1` paths with a `read_dir("/sys/class/drm")` scan — on my machine only `card1` exists, and hybrid laptops can expose both iGPU and dGPU in non-obvious order.
- Narrowed the env-var-set to only the NVIDIA branch. Intel/AMD on Mesa already default to DMA-BUF enabled in current WebKitGTK, so the explicit `=0` was redundant.

Thanks again for digging into this — nice improvement for the Linux-NVIDIA users.

PR #219, opened 2026-04-19
fix(player): stabilize seekbar backward seeking on uncached streams

### LINKED ISSUE #218: BUG: Issue when trying to go backwards in a song using the seekbar
When trying to search in a song using the seekbar, it's sometimes impossible to go backwards. For some reason it's always possible to go forward in the song though. When the issue appears it can be solved by loading the song again.

How to reproduce:
1. Start a song which is not already buffered.
2. As soon as the song starts, go to somewhere close to the ending and try to go backwards immediately after.

I'm using Windows 11

cucadmuh:
Related issue: [#218 — BUG: Issue when trying to go backwards in a song using the seekbar](https://github.com/Psychotoxical/psysonic/issues/218).

Closes #218.

## What changed

- Reworked stream seek recovery logic in `playerStore` for cold/unbuffered track starts.
- Updated `WaveformSeek` interaction flow to commit seeks more predictably and avoid drag/progress race artifacts.
- Added a fast path for stream sources: try direct `audio_seek` first with a short timeout window.
- Added automatic fallback to byte-path restart when direct stream seek does not complete in time.
- Added guarded retry flow after restart so backward seeks are eventually applied even under unstable network/chunk loading.
- Prevented UI/progress regressions during recovery by keeping seek target guards and coalescing repeated seek attempts.

## Why

- Issue #218 reproduces most often right after track start when data is not yet buffered: forward seek may work while backward seek can fail or stick.
- A direct-seek probe keeps interaction snappy when enough data is already available.
- Deterministic fallback keeps behavior stable when the stream path is not yet seekable, without reintroducing freezes and audio artifacts.

## Files

- `src/store/playerStore.ts`

### DIFF
--- a/src/components/WaveformSeek.tsx
+++ b/src/components/WaveformSeek.tsx
@@ -832,6 +832,9 @@ export default function WaveformSeek({ trackId }: Props) {
   useEffect(() => {
     return usePlayerStore.subscribe((state, prev) => {
       if (state.progress === prev.progress && state.buffered === prev.buffered) return;
+      // While user drags, keep the local preview stable. External progress ticks
+      // during streaming/recovery would otherwise fight the cursor and flicker.
+      if (isDragging.current) return;
       progressRef.current = state.progress;
       bufferedRef.current = state.buffered;
       if (!ANIMATED_STYLES.has(styleRef.current)) {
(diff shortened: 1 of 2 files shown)


Psychotoxical:
Hey @cucadmuh — thanks a lot for digging into this one. I've landed the WaveformSeek UX parts of your fix on `main` via commit ccf6e6c8 (with you as co-author):

- Split into `previewFraction` (during drag) / `commitSeek` (on mouseup) so `audio_seek` fires once per gesture instead of on every mousemove.
- Subscribe-effect ignores external progress ticks while `isDragging` — kills the cursor flicker from streaming/recovery ticks fighting the dragged position.
- Cleared `seekTarget` in the `seek()` catch branch so a failed seek can't permanently block progress updates.

I tested it against your  Navidrome from Germany (high latency / lossy upstream) — backward + forward seeking both stay smooth now, no more frozen waveform. 

I left out the byte-restart fast-path with the 450 ms probe + retry state machine on purpose. The reason: there's a parallel branch (`test/seekable-streaming`, also merged in the same FF) that introduces `RangedHttpSource` — a fully seekable HTTP `MediaSource` for streaming starts. With that in place, `audio_seek` succeeds directly and the byte-restart fallback would actually undo the seekable-streaming benefit (we'd throw away the linearly downloaded buffer and start a fresh stream from byte 0 every time the seek didn't complete in 450 ms).

Closing this PR since the user-facing freeze is now fixed on `main`. Huge thanks again for the analysis and the clean diff — it made the cherry-pick trivial.

PR #235, opened 2026-04-21
fix(subsonic): align Rust HTTP UA with main WebView UA

### LINKED ISSUE #233: Bug: different user-agents at the same time
I found that Psysonic is accessing the server using different user agents.

Image

When I find time, I'll try to understand the reason and correct this behavior.

cucadmuh:
Related issue: [#233 — Bug: different user-agents at the same time](https://github.com/Psychotoxical/psysonic/issues/233).

Closes #233.

## What changed

- Added runtime UA sync command in Tauri backend so Rust-side HTTP uses the same UA source as the main WebView.
- Restricted UA updates to the `main` window label to avoid mini-window overwriting backend UA.
- Unified Rust-side HTTP clients in `lib.rs` to use `subsonic_wire_user_agent()` consistently.
- Updated audio engine HTTP client lifecycle in `audio.rs` to refresh UA when runtime UA changes.
- Kept startup sync logic in `main.tsx` minimal: single one-shot sync from main WebView.

## Why

- Navidrome can show multiple simultaneous clients when different UA strings are used by frontend/webview and Rust HTTP paths.
- The audio engine previously held a long-lived HTTP client and could continue using stale UA after runtime updates.
- Using the main WebView UA as the source of truth keeps client identity consistent with platform-native WebView behavior.

## Files

- `src/main.tsx`
- `src-tauri/src/lib.rs`
- `src-tauri/src/audio.rs`
- `src/api/subsonic.ts`

## Attention Before Merge

> [!IMPORTANT]
> Mandatory manual verification on **Windows** and **macOS** is required before merge.
> Confirm that Navidrome no longer shows mixed concurrent user-agents for the same session.

## Open PR

Base: `Psychotoxical/psysonic` → `main`, head: `fix/subsonic-client-headers`.

https://github.com/cucadmuh/psysonic/pull/new/fix/subsonic-client-headers

### DIFF
--- a/src-tauri/src/audio.rs
+++ b/src-tauri/src/audio.rs
@@ -1,5 +1,5 @@
 use std::io::{Cursor, Read, Seek, SeekFrom};
-use std::sync::{Arc, Mutex, OnceLock, TryLockError};
+use std::sync::{Arc, Mutex, OnceLock, RwLock, TryLockError};
 use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, AtomicUsize, Ordering};
 use std::time::{Duration, Instant};
 #[cfg(unix)]
@@ -1980,7 +1980,7 @@ pub struct AudioEngine {
     pub current: Arc<Mutex<AudioCurrent>>,
     /// Monotonically incremented on each audio_play (non-chain) / audio_stop call.
     pub generation: Arc<AtomicU64>,
-    pub http_client: reqwest::Client,
+    pub http_client: Arc<RwLock<reqwest::Client>>,
     pub eq_gains: Arc<[AtomicU32; 10]>,
     pub eq_enabled: Arc<AtomicBool>,
     pub eq_pre_gain: Arc<AtomicU32>,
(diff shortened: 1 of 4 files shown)


Psychotoxical:
### Linux ✅
Tested on Arch Linux (WebKitGTK 6.0, NVIDIA proprietary) against my Navidrome instance:
- **Before:** 2 rows per session — `psysonic [Safari/Linux]` (WebView) + `psysonic [psysonic]` (Rust backend with hardcoded `psysonic/<ver>` UA, Navidrome couldn't parse a platform out of that)
- **After:** 1 row — `psysonic/1.43.0 [Safari/Linux]`, both WebView and Rust HTTP requests now identify as the same client

### Windows ✅
Tested on Windows 11 (WebView2 / Edge):
- **Before:** 2 rows — `psysonic [Edge/Windows]` + `psysonic/1.42.1 [Edge/Windows]` at the same timestamp
- **After:** 1 row — `psysonic/1.43.0 [Edge/Windows]`

### macOS ⏳
I'll test on macOS later today and follow up.

### One scope note
The change also touches UA headers for non-Navidrome clients (`search_radio_browser`, `get_top_radio_stations`, `lastfm_request`, `autoeq_*`, `download_update`, offline/sync downloads), replacing their previous hardcoded `psysonic/1.0` / `psysonic/1.13.0` with the WebView UA. Functionally fine, but radio-browser.info and Last.fm benefit from an `psysonic/<version>` identifier over a generic WebKit UA. Not blocking — could be a small follow-up where we keep `psysonic/<ver>` for external APIs and only align the Navidrome path.

Psychotoxical:
Mac confirmed

PR #237, opened 2026-04-21
feat(logging): add runtime log levels and debug log export

Psychotoxical:
Cherry-picked from @cucadmuh's `fix/ui-freeze-during-streaming` branch. Base branch is #236 so the `eprintln!` → `app_eprintln!` rewrites in `audio.rs` apply cleanly — once #236 merges this will retarget `main` automatically.

## Summary
- New `src-tauri/src/logging.rs` with three levels (Off / Normal / Debug) and a bounded in-memory ring buffer for recent lines.
- `app_eprintln!` macro replaces all direct `eprintln!` calls in `audio.rs`, `cli.rs`, `lib.rs`, `taskbar_win.rs` so they route through the level filter and the ring buffer.
- Settings → System exposes the level dropdown plus an "Export debug log" button (enabled only when Debug is active).
- Local timestamps formatted for readability.
- i18n strings for all 8 locales.

## Test plan
- [x] `cargo check` + `npm run build` green.
- [ ] Toggle levels in Settings, verify filtering works
- [ ] Export debug log, confirm file contents match the ring buffer

Credit: all commits authored by @cucadmuh.


Psychotoxical:
Auto-closed because I squash-merged the base branch (#236) with `--delete-branch`, which invalidated this PR's base. Replaced by #241 with the same commit. Apologies for the noise.

PR #273, opened 2026-04-23
fix(windows): idle WebView2 when Tauri windows are hidden

peri4ko:
### Problem

On **Windows (WebView2)**, hiding a window with Tauri’s `win.hide()` does not always behave like a normal browser tab hide: the native window disappears, but the engine can keep **compositor and main-thread work** going (CSS `@keyframes`, decorative canvas loops, periodic JS timers, etc.). Often **`document.hidden` / `visibilitychange` do not line up** with “the native window is already hidden”, so Page Visibility alone is not enough to pause visual work.

### Scope

Mitigation is **best-effort reduction of visual/compositor and periodic JS wakeups**, not a full “freeze the JS runtime”.

- **`window.__psyHidden`** — set from injected JS when we pause or hide a webview; the UI reads it where `document.hidden` is not enough.
- **Rust → `eval` pause/resume** — sets/clears `__psyHidden`, optional `--psy-anim-speed`, walks **`#root`** descendants to set/clear inline **`animation-play-state`** for **`@keyframes`** (not CSS `transition`, not arbitrary timers / all rAF). **`eval(PAUSE)` runs before `hide()`** on tray, mini, native mini close, and related paths.
- **`data-psy-native-hidden`** on `<html>` from the same inject — drives the same global **`animation-play-state: paused !important`** CSS as **`data-app-hidden`**, so **portaled UI and pseudo-elements** still pause when WebView2 keeps `document.hidden === false` after `win.hide()`.
- **Commands** `pause_rendering` / `resume_rendering` — used from JS on close-to-tray.
- **`WindowVisibilityProvider`** — adaptive polling of `document.hidden` with reliable timer cleanup.
- **`MiniPlayer`** — skips `audio:progress` updates when hidden / `__psyHidden`.
- **`WaveformSeek`** — while hidden, a **~400 ms poll** instead of a tight rAF loop; teardown cancels both.
- **`Hero`** — no 10 s carousel interval while the window is hidden.
- **`PlaybackScheduleBadge`** / **`usePlaybackScheduleRemaining`** — no 400/500 ms intervals while hidden; ticks respect `document.hidden` and `__psyHidden`.

### Limitations

- Does **not** stop **all** timers, **all** rAF, or **all** GPU use — WebView2 can still hold a GPU context; we reduce **content-driven** churn.
- Inline inject still targets **`@keyframes`** under **`#root`** only; global pause for portaled content relies on the **`html[data-psy-native-hidden]`** / **`data-app-hidden`** CSS rules.

### Files

- `src-tauri/src/lib.rs`
- `src/App.tsx`
- `src/components/MiniPlayer.tsx`
- `src/components/Hero.tsx`
- `src/components/PlaybackScheduleBadge.tsx`
- `src/utils/playbackScheduleFormat.ts`
- `src/hooks/useWindowVisibility.tsx`
- `src/styles/components.css`

### Verification

- **Windows:** tray hide/show, mini toggle, pre-created hidden mini, sleep-timer / delayed-start when armed — GPU and periodic spikes should improve; no regression when windows are visible.
- **Linux / macOS:** same flows — behaviour unchanged or safer; relevant paths are not `#[cfg(windows)]`-gated away.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -3039,8 +3039,13 @@ fn default_mini_position(app: &tauri::AppHandle) -> Option<tauri::PhysicalPositi
 /// or every `requestAnimationFrame` loop — it sets a flag the app reads, zeros
 /// `--psy-anim-speed` (for CSS that opts into it), and pauses **@keyframes**
 /// animations via `animation-play-state` (not CSS transitions).
+///
+/// Also sets `data-psy-native-hidden` on `<html>` so global CSS can pause every
+/// animation including `::before`/`::after` and portal content under `<body>`
+/// when `document.hidden` stays false on some WebView2 builds after `win.hide()`.
 const PAUSE_RENDERING_JS: &str = r#"
 window.__psyHidden = true;
+document.documentElement.setAttribute('data-psy-native-hidden', 'true');
 document.documentElement.style.setProperty('--psy-anim-speed', '0');
 (function () {
   const root = document.getElementById('root');
@@ -3054,6 +3059,7 @@ document.documentElement.style.setProperty('--psy-anim-speed', '0');
 /// JS snippet to resume rendering when the window becomes visible again.
 const RESUME_RENDERING_JS: &str = r#"
 window.__psyHidden = false;
+document.documentElement.removeAttribute('data-psy-native-hidden');
 document.documentElement.style.removeProperty('--psy-anim-speed');
 (function () {
   const root = document.getElementById('root');
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -2748,7 +2748,9 @@ fn build_tray_icon(app: &tauri::AppHandle) -> tauri::Result<TrayIcon> {
                 if let Some(win) = app.get_webview_window("main") {
                     if win.is_visible().unwrap_or(false) {
                         let _ = win.hide();
+                        let _ = win.eval(PAUSE_RENDERING_JS);
                     } else {
+                        let _ = win.eval(RESUME_RENDERING_JS);
                         let _ = win.show();
                         let _ = win.set_focus();
                     }
@@ -2782,7 +2784,9 @@ fn build_tray_icon(app: &tauri::AppHandle) -> tauri::Result<TrayIcon> {
                 if let Some(win) = app.get_webview_window("main") {
                     if win.is_visible().unwrap_or(false) {
                         let _ = win.hide();
+                        let _ = win.eval(PAUSE_RENDERING_JS);
                     } else {
+                        let _ = win.eval(RESUME_RENDERING_JS);
                         let _ = win.show();
                         let _ = win.set_focus();
                     }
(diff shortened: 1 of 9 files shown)


cucadmuh:
Thanks for tackling the WebView2 “hidden window still feels alive” issue — pausing CSS animations and gating canvas work behind an explicit `__psyHidden` flag is a sensible direction.

A few notes that might help tighten the change or the description:

**Comments vs. what the injected JS actually does**  
The Rust comment mentions stopping rAF loops and `setInterval` timers. The snippet only sets `__psyHidden`, tweaks `--psy-anim-speed` (currently unused elsewhere in the repo), and pauses **CSS animations** via `animationPlayState`. It does not touch `setInterval` / `clearInterval`, and it does not pause **CSS transitions** (only `@keyframes`-style animations). Consider aligning the comment with the implementation so future readers don’t expect full “JS engine idle.”

**`WaveformSeek` rAF**  
When `document.hidden` or `__psyHidden` is true, `tick` still schedules `requestAnimationFrame(tick)` every frame — you skip `drawSeekbar`, but the rAF loop keeps spinning. If the goal is to minimize compositor wakeups, you could only reschedule rAF when visible again (or use a slow `setTimeout` fallback while hidden), and cancel on unmount as today.

**`querySelectorAll('*')`**  
On a large DOM (especially the main window), running this on every hide/show could cause a noticeable main-thread spike. Worth keeping an eye on profiles; if it shows up, scoping to a root container or a class prefix might be enough.

**`WindowVisibilityProvider` timer cleanup**  
`scheduleCheck` chains `setTimeout` recursively, but only the first timeout id is stored in `timerRef`. On unmount, only that id is cleared, so later timeouts in the chain can still fire (tests, HMR, or any future remount). Storing the latest id on each schedule (or using a single `setInterval` / `requestAnimationFrame` with a cancelled flag) would make cleanup reliable.

**`document.hidden` on WebView2**  
If `win.hide()` doesn’t flip `document.hidden` in some builds, the context-based `hidden` state may drift from `__psyHidden` for components that only use `useWindowVisibility()`. Might be worth a quick check on a Windows machine; if it’s flaky, leaning on `__psyHidden` (or a small Tauri event from Rust when visibility changes) could be more reliable than polling alone.

**Tray hide path**  
You `hide()` then `eval(PAUSE)` — probably fine, but there’s a tiny window where the window is hidden before the flag is set. Not sure it matters in practice; `RESUME` before `show()` is the nicer ordering on the way back.

None of these block appreciation of the approach; they’re mostly about accuracy, edge cleanup, and squeezing a bit more out of the mitigation if GPU is still high after this lands.

peri4ko:
ok, work also on periodically GPU load, wait, pls

peri4ko:
all what i want to do is done

cucadmuh:
Thanks for the follow-up commit — `0e07a73` lines up well with the review: accurate rustdoc, `#root`-scoped animation pause, eval-before-hide, WaveformSeek throttling, and fixed `WindowVisibilityProvider` cleanup.

A few notes: the opening PR description is now a bit out of date (pause runs **before** `hide()`, and the snippet doesn’t “halt all rendering” / all timers). Worth editing the top post so it matches the code.

For the periodic GPU spikes you mentioned: the 10 s **Hero** carousel on Home is a likely culprit; we’re also looking at pausing it when the window is Tauri-hidden (WebView2 doesn’t always flip `document.hidden`) and at gating the sleep-timer badge intervals — can land as a small follow-up after this merges (or a separate PR if you prefer to keep #273 tight).

cucadmuh (approved):

Psychotoxical (approved):
Tested on macOS — all flows work as expected (mini close, tray toggle, minimize/restore, Cmd+W).

A few non-blocking notes:
- `src/App.tsx:934` has a stray leading-space indent on the `// window:close-requested` comment.
- `(window as any).__psyHidden` could be typed once via `declare global { interface Window { __psyHidden?: boolean } }` to drop the casts.
- `WindowVisibilityProvider` only polls `document.hidden`. On the WebView2 path the Rust-injected `__psyHidden` is the more reliable signal — OR-ing it in would let the `if (windowHidden) return` effect guards actually skip creating intervals, instead of relying on the per-tick fallback check.
- 200 ms visible-polling is a bit aggressive (5 wakeups/s just for visibility) — 500 ms would probably be fine.

LGTM otherwise.

PR #281, opened 2026-04-23
perf(ui): fix spike when medulla-perch lines up with hair-fan gestures

cucadmuh:
The harness parks the pointer on the micro-target medulla perch, then runs a hair-fan sweep through it; that combination used to restyle and repaint more of the chrome than the compositor could absorb. Localize vector fill identity per glyph instance, keep full-window dimming in a bounded surface, and trim the Settings subtree path so layout and paint work cannot stack on the same frame.


Psychotoxical (approved):
Reviewed the localised vector-fill identity per glyph instance and the bounded full-window dimming surface — both behave as described, no observable restyle storm in the relevant frame budget. LGTM.

Psychotoxical:
Reverted in #282 — follow-up incoming with the chrome restyle path bounded more tightly.

PR #283, opened 2026-04-23
perf(ui): fix spike when medulla-perch lines up with hair-fan gestures

cucadmuh:
The harness parks the pointer on the micro-target medulla perch, then runs a hair-fan sweep through it; that combination used to restyle and repaint more of the chrome than the compositor could absorb. Keep full-window dimming on a bounded surface and trim the Settings subtree so layout and paint work cannot stack on the same frame. When the modal perch drew a second wordmark beside the sidebar roost, both must not share one `psysonicGrad` nest—thread a disposable suffix through the portal perch only. Leave the export shortcut inscription and its trailhead on the main-branch rune so unrelated shell copy stays byte-for-byte with upstream.


Psychotoxical (approved):
Reviewed locally — the bounded full-window dimming surface holds, the Settings subtree trim avoids the layout/paint frame stacking, and threading the disposable suffix through the portal perch keeps the second wordmark's gradient identity isolated from the sidebar roost. LGTM.

PR #326, opened 2026-04-26
fix(analysis,waveform): stabilize LUFS loop, serialize CPU seed work, and introduce mixed mean/peak waveform bins

cucadmuh:
## Summary (branch theme)

This branch finalizes a coordinated stabilization pass over the new analysis stack:

1. It removes a loudness feedback loop that could freeze the Windows UI under active LUFS normalization.
2. It serializes full-track waveform/loudness seeding through a dedicated CPU queue, eliminating duplicate heavy decodes and duplicate update emits.
3. It upgrades waveform cache/output to a dual-curve model (`peak` + `mean(abs)` per bin) and renders with a fixed 70/30 mix.

Together these changes make waveform + loudness behavior more deterministic under streaming, gapless transitions, and cache-miss playback.

## Why

- **UI responsiveness risk:** periodic partial-loudness updates could trigger repetitive replay-gain state churn and overload the renderer event path.
- **Wasted CPU during cache miss:** multiple producers could request full analysis for the same track close in time, causing repeated expensive decode work.
- **Waveform shape quality:** single-curve peak-only bins are stable but can look sparse/spiky; mean(abs) contributes denser contour information.
- **Cache safety during format evolution:** once waveform payload shape changes, stale/insufficient rows must not be silently reused.

### 1) Loudness feedback-loop stabilization

- Tightened frontend loudness refresh/update sequencing in `playerStore` so normalization updates do not recursively flood control paths.
- Added backend-side emit dedupe and safer emission boundaries so loudness and normalization state events are not spammed under steady values.
- Result: the Windows freeze scenario reported under sustained LUFS partial updates is removed.

### 2) Full-analysis CPU seed queue (single worker, prioritized)

- Introduced a dedicated analysis CPU queue in `src-tauri/src/lib.rs`:
  - full `seed_from_bytes` work is submitted centrally,
  - high-priority requests (current track) are promoted,
  - heavy decode work runs serialized in one worker lane.
- Moved event emission responsibility to queue completion path:
  - `analysis:waveform-updated` is emitted once from submit/complete path,
  - duplicate emits from lower-level call sites were removed.
- Added periodic queue snapshot logging in debug mode for observability (startup line + interval snapshots).

### 3) Duplicate-seed suppression and pipeline cleanup

- Refactored waveform/loudness seed funnel in `analysis_cache` and `audio` integration paths to avoid overlapping decode requests for identical effective work.
- Cleaned code paths around concurrent triggers (stream completion, preload bytes, and explicit backfill), reducing redundant CPU pressure.

### 4) Waveform v4: dual-curve bins + strict cache shape validation

- Bumped waveform algorithm version in `analysis_cache` to **v4**.
- Extended PCM scan aggregation to compute, per time bin:
  - max magnitude (`peak`),
  - mean absolute magnitude (`mean(abs)`).
- Persisted bins as a paired payload: first half `peak[500]`, second half `mean[500]`.
- Added explicit cache-row sufficiency check (`waveform_cache_blob_len_ok`):
  - expected shape is `2 * bin_count`,
  - malformed/insufficient rows are ignored (treated as cache miss) and naturally re-seeded.
- Byte-fallback path now mirrors v4 shape by duplicating peak bins into both halves for compatibility.

### 5) Frontend waveform rendering mix

- Updated `WaveformSeek` conversion path:
  - when payload length is `1000` (v4), it applies fixed blend
    - `0.7 * mean + 0.3 * peak`;
  - when legacy `500` bins are seen, it falls back to old single-curve behavior.
- Updated waveform payload coercion/validation in `playerStore`:
  - accepts only known valid lengths (`500` or `1000`),
  - rejects malformed values early to prevent rendering corruption.

### 6) Logging and diagnostics polish

- Added URL-safe logging utility for Subsonic resources to avoid leaking sensitive URL details in debug output.
- Retained actionable analysis diagnostics (queue depth, seed outcomes, decode path hints) while reducing noisy/duplicate event logs.

## Commit coverage (main..branch)

- `2d222e2` fix(loudness): break feedback loop that froze Windows UI when LUFS active
- `5cb233e` fix(analysis): dedupe concurrent seed_from_bytes for the same track_id
- `218aa00` fix(analysis): CPU seed queue, single waveform emit, and log URL redaction
- `8f603e8` feat(waveform): store peak+mean bins and blend at 70/30

## Key files

- Backend core:
  - `src-tauri/src/analysis_cache.rs`
  - `src-tauri/src/audio.rs`
  - `src-tauri/src/lib.rs`
- Frontend core:
  - `src/store/playerStore.ts`
- Supporting:
  - `src/hotCachePrefetch.ts`
  - `src/store/hotCacheStore.ts`
  - `src/utils/redactSubsonicUrl.ts`

## Behavior impact (expected)

- Lower risk of renderer stalls under active LUFS updates.
- Less duplicated full-track decode work on cache-miss playback.
- Single, predictable waveform-updated event per completed seed path.
- More visually stable waveform contour from mean/peak blend.
- Safe auto-invalidation/reseed when cached waveform payload is structurally insufficient.

## Risks / trade-offs

- Waveform v4 uses larger per-track payload (`1000` bins vs `500`), slightly increasing cache storage.
- Existing waveform cache rows from older format are intentionally bypassed/reseeded (one-time recomputation cost).
- Mixed-mode period is handled (500 fallback), but visual shape may differ across tracks until reseed converges.

## Acknowledgements

Special thanks to [@peri4ko](https://github.com/peri4ko) for proposing the mean/peak waveform blending direction that informed the v4 waveform shape update.

### DIFF
--- a/src-tauri/src/analysis_cache.rs
+++ b/src-tauri/src/analysis_cache.rs
@@ -14,8 +14,17 @@ use symphonia::core::meta::MetadataOptions;
 use symphonia::core::probe::Hint;
 use tauri::Manager;
 
-pub const WAVEFORM_ALGO_VERSION: i64 = 3;
+pub const WAVEFORM_ALGO_VERSION: i64 = 4;
 pub const LOUDNESS_ALGO_VERSION: i64 = 1;
+
+/// Bins in waveform BLOB: `2 * bin_count` bytes (peak u8, then mean-abs u8 per time bin).
+pub fn waveform_cache_blob_len_ok(bins: &[u8], bin_count: i64) -> bool {
+    if bin_count <= 0 {
+        return false;
+    }
+    let n = bin_count as usize;
+    bins.len() == n.saturating_mul(2)
+}
 
 #[derive(Debug, Clone)]
 pub struct TrackKey {
@@ -200,8 +209,9 @@ impl AnalysisCache {
 
     pub fn get_waveform(&self, key: &TrackKey) -> Result<Option<WaveformEntry>, String> {
         let conn = self.conn.lock().map_err(|_| "analysis_cache lock poisoned".to_string())?;
-        conn.query_row(
-            r#"
+        let row = conn
+            .query_row(
+                r#"
             SELECT w.bins, w.bin_count, w.is_partial, w.known_until_sec, w.duration_sec, w.updated_at
             FROM waveform_cache w
             JOIN analysis_track a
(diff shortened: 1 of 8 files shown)


Psychotoxical (commented):
Works without any problems. Ready to merge

PR #349, opened 2026-04-28
ci(release): gate promote-main on explicit main checks

cucadmuh:
Add an always-on ci-main workflow that produces a stable check name, and validate that check via check runs (not branch protection APIs). Wire promote-main-to-next to require ci-main / ci-ok and document the required check in the release SOP.

### DIFF
--- a/.github/workflows/ci-main.yml
+++ b/.github/workflows/ci-main.yml
@@ -1,0 +1,15 @@
+name: ci-main
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  ci-ok:
+    runs-on: ubuntu-latest
+    steps:
+      - name: ci sentinel
+        run: echo "main CI sentinel is green"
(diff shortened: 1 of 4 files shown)


Psychotoxical (commented):
Looks good — design matches what we agreed on, and dropping the branch-protection API call is the right move.

One thing to double-check before we test promote: the GitHub API's `check_runs[].name` for a standalone workflow is usually just the job name (here: `ci-ok`), not the `<Workflow> / <Job>` format the UI shows. The `validate-main / validate` we saw in earlier logs was specifically because `validate-main` is a reusable workflow (`workflow_call`). For a regular push-triggered workflow like `ci-main`, the API name might just be `ci-ok`.

Easiest way to confirm: after this merges, `ci-main` runs once on main — just check the API name via:

```
gh api repos/Psychotoxical/psysonic/commits/main/check-runs --jq '.check_runs[].name'
```

If it returns `ci-ok` instead of `ci-main / ci-ok`, the `required_contexts` default needs adjusting. Otherwise we're good to go.

Minor nit: `permissions.statuses: read` can be dropped now since the script no longer reads statuses. Cosmetic, not blocking.

PR #350, opened 2026-04-28
ci(release): run channel publish after promote when GITHUB_TOKEN push skips push rules

cucadmuh:
GitHub does not fire push workflows for commits pushed with the default actions token. Chain Next/Release channel workflows from the matching promote workflow_run on success.

Made-with: Cursor

### DIFF
--- a/.github/workflows/next.yml
+++ b/.github/workflows/next.yml
@@ -1,16 +1,23 @@
 name: Next Channel
 
+# Pushes made with the default GITHUB_TOKEN from another workflow (e.g. Promote main to next)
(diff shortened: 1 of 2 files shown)


Psychotoxical (approved):
LGTM — addresses a real GitHub Actions limitation (default GITHUB_TOKEN pushes don't fire downstream workflows). The `workflow_run` chain + `source_ref` ternary for the default-branch context is the right shape, and the `if: conclusion == 'success'` guard prevents builds after a failed promote. Inline comments explain the why nicely.

PR #419, opened 2026-05-01
feat: queue-ux-improvements

kveld9:
### Overview
This PR introduces three major UX improvements to the Queue Panel: a collapsible "Now Playing" section, an estimated time of arrival (ETA) display in the header, and a visual equalizer indicator for the currently playing track.

---

#### `src/components/QueuePanel.tsx`
**Changes:**
- **Collapsible Queue Header**: Added `isNowPlayingCollapsed` state with a chevron button to toggle visibility of the "Now Playing" section and toolbar icons
- **ETA Display**: Added estimated finish time calculation showing when the queue will end
  - Uses `useMemo` for `futureTracksDuration` to avoid recalculating on every tick
  - Throttles store subscription to 30-second precision (`Math.floor(currentTime / 30) * 30`)
  - Dynamic styling: full opacity with accent color when playing, muted when paused
- **Track Counter**: Changed format from "X tracks" to "(current/total)" format: `(24/102)`
- **Visual Equalizer**: Added animated EQ bars to the left of the currently playing track in the queue list using existing `.eq-bars` CSS classes
- **Code Quality**: Removed dead code (`totalSecs`, `fmt` function), added proper TypeScript typing (`Track` instead of `any`)

#### `src/locales/*.ts` (8 files)
**Added translation keys:**
- `queue.hideNowPlaying` / `queue.showNowPlaying` - Tooltip text for collapse button

**Files updated:** es.ts, en.ts, zh.ts, ru.ts, nl.ts, de.ts, fr.ts, nb.ts

---

### Technical Details

**ETA Calculation Logic:**

```typescript
const currentTime = usePlayerStore((s) => Math.floor(s.currentTime / 30) * 30);
const isPlaying = usePlayerStore((s) => s.isPlaying);

const futureTracksDuration = useMemo(() =>
  queue.slice(queueIndex + 1).reduce((acc, track) => acc + (track.duration || 0), 0),
  [queue, queueIndex]
);

const remainingSecs = Math.max(0,
  (queue[queueIndex]?.duration ?? 0) - currentTime + futureTracksDuration
);

const eta = new Date(Date.now() + remainingSecs * 1000)
  .toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
```

**Performance Optimizations:**
- `useMemo` prevents O(n) reduce operation on every render
- Flooring `currentTime` to 30-second buckets reduces re-renders from ~1/sec to ~2/min
- Removed redundant calculations and dead code

**Visual States:**

| State | Counter | ETA Color | ETA Opacity |
|-------|---------|-----------|-------------|
| Playing | `var(--text-muted)` | `var(--accent)` | 1.0 |
| Paused | `var(--text-muted)` | `var(--text-muted)` | 0.5 |

---

### Result

**Header:** `Queue (24/102) · 02:10`
- ETA prominently visible with accent color during playback
- Updates every 30 seconds (sufficient for minute-level precision)
- One-click collapse/expand for queue toolbar and now playing info
- Active track shows animated equalizer bars on the left


Psychotoxical (commented):
Tested locally on Linux. Nice quality-of-life additions overall — the `(N/M)` counter and the ETA pull a lot of weight, and the 30-second bucketing is a clean way to keep the re-renders down.

Four small things came up while clicking through it:

### 1. The ETA in the header is not self-explanatory

`Queue (24/102) · 02:10` doesn't immediately read as "queue ends at 02:10" — I had to stop and figure it out. A short tooltip would close that gap. The project already has a global `TooltipPortal` (in `App.tsx`) driven by `data-tooltip="…"` attributes, so it's a one-liner on the span around `QueuePanel.tsx:228` plus a single new i18n key (e.g. `queue.etaTooltip` = "Estimated end of queue").

### 2. `isNowPlayingCollapsed` doesn't survive a restart

`QueuePanel.tsx:342` uses `useState(false)`, so the collapsed/expanded state resets every mount. Convention in this codebase is to put persistent UI toggles on `authStore` (cf. `minimizeToTray`, `showFullscreenLyrics`, `lyricsServerFirst`, `enableAppleMusicCoversDiscord`). A `queueNowPlayingCollapsed: boolean` field there would match that pattern.

### 3. EQ bars keep animating during pause

The `.eq-bar` CSS animation runs unconditionally for the active track, so when you hit Pause the bars keep dancing as if the song were still playing. Easiest fix is to gate the animation on the store's actual `isPlaying`:

```tsx
const isStorePlaying = usePlayerStore((s) => s.isPlaying);
…
{isPlaying && (
  <div className={`eq-bars${isStorePlaying ? '' : ' paused'}`} …>
    …
  </div>
)}
```

```css
.eq-bars.paused .eq-bar { animation-play-state: paused; }
```

### 4. Redundant `<Play>` icon next to the title (pre-existing — but now duplicated)

`QueuePanel.tsx:1015` renders `<Play size={10} />` next to the active track's title. That predates this PR (it was the original active-track indicator), but now the EQ bars play exactly that role on the left side of the row, so the same track gets two "this is the playing one" markers. Suggest dropping line 1015 in this PR — the EQ bars are the better signal.

Happy to push the fixes myself if that's easier — otherwise feel free to take this as a checklist.

kveld9:
Hi, thank you for the corrections. I've pushed a commit trying to fix all of that; at least from what I've tested, it seems to be working fine. I'm here if you need anything.

Psychotoxical:
Pushed `ae68139` as a maintainer edit — hope you don't mind. Quick rationale:

The standalone ETA span next to the position counter collided with the existing total/remaining time toggle in the header (the clickable label that lets you flip between `2:35:12` and `-1:25:30`). Rather than lose that toggle or run two duration displays side-by-side, I folded ETA in as a third mode: clicking the label now rotates `total → remaining → eta → total`. Your live-feel treatment for ETA (accent colour while playing, muted at 50% opacity when paused) is preserved on the eta mode only.

Counter `(N/M)`, collapsible Now Playing section, persistence via authStore, EQ indicator + pause, removed redundant Play icon — all of that stays as you had it. Nice catch on the hook-order issue too (the `if (queue.length === 0) return null;` before `useMemo`), that was a latent bug.

i18n: dropped `queue.etaTooltip` (no longer needed as a descriptive label since ETA is now an action of the toggle), added `queue.showEta` across all 8 locales to match the existing `showRemaining` / `showTotal` action-tooltip pattern.

Thanks for the PR — all three improvements are great. Ready to merge from my side.

PR #463, opened 2026-05-05
Environment upgrade & hot-cache playback

cucadmuh:
**Branch:** `chore/full-environment-upgrade`  
**Base:** `main`

## Summary

Dependency and tooling refresh (frontend + Tauri/rust, **rodio 0.22**), smaller bundle routing, two focused UI fixes (Windows tray, sidebar “new releases” persistence), and a larger **streaming / hot-cache** pass so repeat listens and end-of-queue replay prefer **RAM or disk** instead of full HTTP ranged downloads when data is already available.

## User-visible

- **Replay / play again:** After a full listen—or queue end with `audio_stop`—starting the same track again can use **hot cache** or in-memory completed stream instead of re-downloading the whole file over HTTP (when hot cache is enabled and promotion succeeds).
- **Player chrome:** Cache vs stream vs offline **source icon** stays correct after **resume**, **queue undo restore**, and **gapless** track switch—not only after `playTrack`.
- **Sidebar:** “New releases” read state no longer “resurrects” incorrect unread badges when the persisted ID list hits the **500-id cap** (fresh reads merged in, capped set stays coherent).
- **Windows:** Tray **double-click** opens the window without an unintended context menu flash.
- **Build:** Heavier pages are **lazy-loaded**; default Vite **chunk size** warnings are enabled again for visibility.

### Audio / Rust

- **`stream_completed_cache`** is **not cleared** on `audio_stop`, so a completed ranged buffer is not thrown away when the queue ends—aligned with replay and promote-to-hot flows.
- **Same-track** `playTrack` and **cold `resume`** (after `audio:ended`, `isAudioPaused` false) **await / run promote** so `resolvePlaybackUrl` can switch to `psysonic-local://` before `audio_play`.
- **`fetch_data`** still **takes** the completed-stream slot once; the above promotion paths avoid relying on a second HTTP pass when disk index is ready.
- **Format hints** for ranged HTTP: merge URL suffix, `Content-Type`, `Content-Disposition`, Subsonic **`song.suffix`** (`streamFormatSuffix` on `audio_play`); optional **small Range probe** only when hints are still missing; avoid treating generic **`video/mp4`** as an audio hint.
- **Analysis noise / CPU:** `cpu_seed_redundant_for_track` skips unnecessary full-file seeds when waveform **and** loudness exist; **`analysis:waveform-updated`** emits only on real DB **`Upserted`**; **gen** check after `await` before writing ranged/legacy promote slots; **`Manager`** in scope for `try_state` on local-file seed skip path.
- **Edge hardening:** `.catch` on the same-track **promote → `runPlayTrackBody`** chain (gen guard, prefetch defer reset, stop optimistic play).

### Frontend

- **`playerStore`:** promote/bind fixes, `activeServerId` **narrowing** for `tsc`, playback source hints shared helper.
- **`hotCacheStore`:** simplified debug logging (direct `authStore` import instead of dynamic import) for hot-cache events.
- **`App.tsx` / routing:** lazy route boundaries; related **ContextMenu** / **vite** adjustments as in the branch.

### Dependencies

- Broad **npm** and **Cargo** updates with **rodio 0.22** migration and decode/engine glue changes across the audio stack.

## Commits (branch vs `main`, newest first)

| Commit   | Topic |
|----------|--------|
| `a9d5d2c` | same-track promote promise `.catch` |
| `1acf609` | TS: narrow `activeServerId` for promote |
| `417f5ea` | hot-cache replay, analysis skips, playback source UI, streaming/hints |
| `eea1e67` | sidebar new-releases cap merge |
| `8bed9fe` | Windows tray double-click |
| `7380372` | lazy routes + Vite chunk warnings |
| `4fd98b8` | deps + rodio 0.22 |

## Suggested testing

- Stream a track until ranged buffer completes → **stop at queue end** or same track → **Play** / **resume**: expect **local file** or **fetch_data** RAM path, not full ranged re-download (logs: `LocalFileSource` or no repeated full `dl progress`).
- **Hot cache on/off**, **repeat-one**, **same row play** again: icon matches source (waves vs disk vs folder).
- **New releases** with many read IDs: cap behavior, badge consistency.
- **Windows:** tray double-click.
- **`npm run build`** / **`tauri build`** (production `tsc` path).


Psychotoxical (approved):
Solid pass — branch is clean. From my side:

- Hot-cache replay/resume: same-track promote → runPlayTrackBody with gen-guard + .catch is the right shape. Keeping stream_completed_cache alive across audio_stop is exactly the pattern needed for replay / queue-end resume.
- Format-hint pipeline (Content-Type → Disposition → Subsonic suffix → small Range probe) is nicely tiered; the video/mp4 filter is a good detail catch.
- cpu_seed_redundant_for_track closes the analysis no-op gap well.
- Sidebar 500-id cap merge: small but annoying bug, cleanly fixed.
- rodio 0.22 migration + lazy routes go through mechanically.

Audit found zero bottlenecks outside the hot-cache scope. 👍

PR #471, opened 2026-05-05
feat(artist): group albums by release type on artist page

Sayykii:
Uses the releaseType field to group albums/releases into sections like Albums, Compilation, Live, etc. If there's no release type it falls back to normal view

No release type: 
image

With different release types:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * **Live search** and the **mobile search overlay** now load **artist images** for rows in the **Artists** section via the same **`getCoverArt` / image-cache path** as album art (**`coverArt`** when present, otherwise the **artist id** where the server supports it), with a **fallback icon** when art is missing or fails.
 * **Mobile** artist hits use a **round** thumbnail next to **square** album art so the two result types read clearly at a glance.
 
+### Artist page — group albums by release type
+
+**By [@Sayykii](https://github.com/Sayykii), PR [#471](https://github.com/Psychotoxical/psysonic/pull/471)**
+
+* Albums on the artist page can now be **grouped into sections** by their OpenSubsonic `releaseTypes` (Album, EP, Single, Compilation, Live, Soundtrack, Remix). Section order is deterministic across languages, with unknown types appended at the end.
+* Falls back to the previous flat list when the server doesn't return `releaseTypes` or all albums share the default Album type — no behaviour change for non-OpenSubsonic servers.
+* Section headers are localised in all 8 supported languages.
+
 ## Changed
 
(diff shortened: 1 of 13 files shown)


Psychotoxical (commented):
Hey Sayykii — love this one! It's exactly the kind of polish that makes the artist page feel structured instead of one long wall of covers, especially for artists with a real discography. Tested it locally and the behaviour is solid: groups cleanly when `releaseTypes` is present, falls back to the flat list when it isn't.

Four small things I'd like to polish before we merge — none of them are blockers, just want to land it right:

**1. Release-type labels need i18n**

The labels come from the raw MusicBrainz tag title-cased (`"album"` → `"Album"`), so users on the German / French / Russian / Chinese / etc. UI still see English headers. Could you wire them through `i18next` — e.g. `t('artistDetail.releaseTypes.album')`, with a fallback to the title-cased raw tag for unknown types? Happy to help land the matching keys across `src/i18n.ts` and the per-locale files in `src/locales/` once the lookup is in place.

**2. Group order is non-deterministic**

`Map` preserves insertion order, so the section order ends up depending on the order the server returns the albums in. Would be nicer to sort the groups by a fixed semantic order — e.g. `Album → EP → Single → Compilation → Live → Soundtrack → Remix → Other` — with unknown types at the end alphabetically.

**3. Inline styles → CSS classes**

The repo convention is theming via CSS classes / variables, since inline values can override theme tweaks. The header / count-badge / wrapper would be more theme-friendly as e.g. `.artist-release-group`, `.artist-release-group__header`, `.artist-release-group__count` (in the matching `.css` file). The `var(--text-muted)` and `var(--bg-card)` references carry over as-is.

**4. Multi-type grouping — quick design question**

An album tagged `["album", "live"]` ends up in its own `"Album · Live"` group, separate from albums tagged just `["album"]`. That can produce a few small mixed groups. Did you intend that, or would prioritising work better — e.g. if `live` is present put it in the Live group, else `compilation` → Compilation, else Album? Either approach works, just want to make sure it matches what you had in mind.

Thanks again — and seriously, thanks for being so active on Discord. Ideas like this one are exactly what keeps the project moving 🙌

Sayykii:
Hey, will take a look and fix issues 1 to 3, as for 4 this is the intended behavior or at least it's what makes the most sense to me.

Psychotoxical (approved):
Rebased onto current main, i18n release-type labels filled in across the remaining 7 locales (de, fr, nl, zh, nb, ru, es), CHANGELOG and Contributors entries added.

Diff is purely additive: the new `useMemo` adds the grouping logic, and the single-group fallback in the render path preserves the prior flat-list output bit-for-bit. No lock-file changes, no workflow changes, no touched files outside the expected set. Type-check clean across all eight locales.

Approved.

PR #476, opened 2026-05-06
fix: stabilize preview seekbar, post-sleep audio recovery, and card hover behavior

cucadmuh:
## Summary

This branch fixes three user-facing regressions introduced during recent dependency/runtime updates:

1. **Seekbar drift/jump around preview playback** (main seekbar moving while preview pauses main sink, then jumping after preview end).
2. **Audio output loss after OS sleep/resume** (especially Windows; Linux path also covered) requiring app restart.
3. **Hover-induced card instability** (card lift jitter/pulsation and abrupt artwork zoom behavior).

The final state includes platform resume hooks (Windows + Linux), a guarded watchdog fallback with diagnostics, and UI simplifications that remove card lift while preserving smooth cover zoom.

## User-visible changes

- **Preview behavior:** main player seekbar stays stable during track preview and does not jump forward when preview ends.
- **Sleep/resume reliability:** app attempts to recover output stream automatically after wake-up (Windows via power notifications, Linux via logind signal).
- **Card hover feel:** album/artist cards no longer shift vertically on hover; artwork zoom remains but with smooth timing.
- **Debuggability:** CLI/runtime logs now show watchdog arm/clear/trigger/success paths for audio recovery diagnosis.

### Player / seekbar

- `WaveformSeek` interpolation loop is disabled while preview is active (`previewingId != null`) to avoid extrapolating progress while Rust main sink is paused.
- Added anchor reset on preview freeze-state transition so interpolation restarts from current timestamp and cannot "overshoot then snap back" after preview end.

### Audio recovery after sleep

- Added shared helper module `audio/power_resume.rs`:
  - resume-event debounce,
  - delayed reopen routine using current selected output device,
  - centralized reopen path through existing stream reopen flow.

- Added Windows hook `audio/power_notify_win.rs`:
  - `PowerRegisterSuspendResumeNotification` callback for resume events,
  - triggers async output reopen via shared helper.

- Added Linux hook `audio/power_notify_linux.rs`:
  - subscribes to `org.freedesktop.login1.Manager.PrepareForSleep` on system D-Bus,
  - on wake (`start == false`) triggers shared reopen helper.

- Added unified registration entrypoint in `audio/mod.rs`:
  - `register_post_sleep_audio_recovery(app)` routes by target OS,
  - explicit comment documents why macOS intentionally falls back to generic device watcher path for now.

### Watchdog hardening + diagnostics

- Existing stall watchdog in `device_watcher` was refined to reduce false positives:
  - watchdog now arms only after a long poll gap (>= 15s), interpreted as likely sleep/resume,
  - armed window is time-limited (120s),
  - stall trigger still requires flat sample counter while sink is active, with cooldown.

- Added detailed logs for:
  - watchdog arm/disarm,
  - stall candidate start/clear,
  - reopen trigger and outcome.

### UI hover behavior

- Removed vertical hover lift from key card classes (`album-card`, `artist-card`, base `.card`) to eliminate pointer-edge jitter.
- Removed per-card GPU hints (`translateZ(0)`, `will-change`, backface hints) on artwork images to avoid compositing overhead on software-composited Linux paths.
- `CachedImage` no longer forces inline `transition` (which previously overrode CSS transform timing), so zoom transition timing is now controlled by stylesheet and remains smooth.

## Commits (oldest -> newest)

- `b5f1b60` — fix(player): freeze main seekbar during track preview
- `c56b21b` — fix(audio): recover output after sleep and stalled streams
- `d5f5c87` — fix(ui): remove card hover lift and smooth artwork zoom
- `67276bc` — fix(player): prevent seekbar jump after preview ends
- `3133103` — fix(audio): reduce false watchdog recoveries and add diagnostics
- `81c458b` — chore(ui): drop card GPU hints and clarify macOS sleep scope
- `504d77f` — docs(audio): document intentional Win32 callback pointer lifetime

## Diff footprint vs `main`

- **11 files changed**, **+435 / -53**
- Core files:
  - `src/components/CachedImage.tsx`
  - `src/styles/components.css`
  - `src/styles/theme.css`
  - `src-tauri/src/audio/device_watcher.rs`
  - `src-tauri/src/audio/power_resume.rs`
  - `src-tauri/src/audio/power_notify_win.rs`
  - `src-tauri/src/audio/power_notify_linux.rs`
  - `src-tauri/src/audio/mod.rs`
  - `src-tauri/src/lib.rs`
  - `src-tauri/Cargo.toml`

## Suggested testing

- **Preview seekbar:**
  - start track, start preview, stop/natural-end preview,
  - verify no forward jump and no snap-back of main seekbar.

- **Sleep/resume (Windows):**
  - play track, sleep machine, wake,
  - verify output recovers without app restart,
  - check logs for power callback + reopen result.

- **Sleep/resume (Linux):**
  - same scenario with logind session,
  - verify `PrepareForSleep` wake path and reopen behavior.

- **Watchdog false-positive check:**
  - normal long playback without sleep/resume should not trigger reopen,
  - inspect logs to confirm watchdog remains disarmed unless long poll gap occurs.

- **UI card hover:**
  - verify no card vertical movement/pulsation,
  - verify artwork zoom is smooth and not abrupt.

## Notes / follow-ups

- macOS native resume hook is intentionally out-of-scope in this branch.
- Windows correctness should be validated in CI/runtime because local Windows hardware was unavailable during implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * **Double-click** on the tray icon opens (or focuses) the main window without a spurious **context-menu** interaction; tray module import cleanup.
 
+### Playback stability — preview seekbar, sleep/wake recovery, and card-hover jitter
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#476](https://github.com/Psychotoxical/psysonic/pull/476)**
+
+* **Preview seekbar:** while Rust preview playback pauses the main sink, the main seekbar no longer extrapolates forward, and preview end no longer causes a brief forward jump before snapping back.
+* **Sleep/wake audio recovery:** added post-sleep output reopen hooks for **Windows** (power notifications) and **Linux** (logind `PrepareForSleep` wake signal), plus a guarded fallback watchdog path and richer runtime diagnostics.
+* **False-positive mitigation:** the watchdog now arms only after a long poll gap (sleep/resume-like condition) and logs arm/clear/trigger decisions, reducing unexpected stream reopens during normal playback.
+* **Card hover stability:** removed vertical lift on album/artist/base cards to avoid pointer-edge pulsation, kept artwork zoom smooth, and dropped per-card GPU layer hints that could regress software-composited Linux paths.
+
 ## [1.45.0] - 2026-05-04
 
 ## Added
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -67,6 +67,7 @@ windows = { version = "0.62", features = [
     "Win32_Graphics",
     "Win32_Graphics_Gdi",
     "Win32_System_Com",
+    "Win32_System_Power",
     "Win32_System_Threading",
     "Win32_UI_Controls",
     "Win32_UI_Shell",
--- a/src-tauri/src/audio/device_watcher.rs
+++ b/src-tauri/src/audio/device_watcher.rs
@@ -4,18 +4,79 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use tauri::Emitter;
+use tauri::Manager;
 
 use super::engine::AudioEngine;
 #[cfg(not(target_os = "linux"))]
 use super::dev_io::output_enumeration_includes_pinned;
 
+/// What to tell the frontend after a successful stream reopen.
+pub(crate) enum ReopenNotify {
+    /// Normal path — same as `audio_set_device`.
+    DeviceChanged,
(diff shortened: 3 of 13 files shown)


Psychotoxical (approved):
All five review points addressed cleanly — particularly nice: the watchdog rework (event-armed via 15 s poll-gap detection is much more robust than threshold tweaks), and both `Box::into_raw` sites now carry rationale comments.

Heads-up that the only CI check that runs here is the `ci-ok` sentinel job — actual cargo / Tauri builds only get exercised at release-tag time. Worth a local `npm run tauri:build` before tagging if you can, since this branch touches Win32 power APIs that aren't validated by the sentinel.

LGTM 🙌

PR #488, opened 2026-05-06
perf(tracklist): drop now-playing pulse + EQ-bar animations

Psychotoxical:
## Summary

The currently-playing track in any tracklist was rendered with two animations layered on top of an already-busy DOM:

- `.track-row.active` ran `track-pulse` 3s `opacity` 1 → 0.6 → 1 infinite over the **entire row subtree** (title button, several Lucide icons, star-rating SVGs, hover affordances). On WebKitGTK without compositing — Linux + NVIDIA proprietary driver + `WEBKIT_DISABLE_COMPOSITING_MODE=1` — every animated row falls back to a **full software repaint** of the subtree per frame.
- The "now playing" indicator was three `<span class="eq-bar">` siblings with `transform: scaleY()` `eq-bounce` keyframes — same software-composite story, three per active row.

Symptom: WebProcess pegged at ~80 % CPU and 1.2 GB RSS while a track was playing on **AlbumDetail**, dropped immediately on pause/stop. Composers / Settings / Home (no track rows) were unaffected. Profiler showed continuous Layout & Rendering work in lockstep with `isPlaying`.

## Changes

- `.track-row.active` keeps the `--accent-dim` background, drops the pulse animation.
- The "now playing" indicator becomes a single Lucide `AudioLines` icon (four vertical bars of different heights — reads as an EQ icon, one SVG per active row instead of three animated spans). Used in **AlbumDetail**, **ArtistDetail**, **PlaylistDetail**, **Favorites**, **RandomMix**.
- `.eq-bars` className collapsed to `color: var(--accent); display: block`.
- Cleanup: dead `@keyframes track-pulse`, `@keyframes eq-bounce`, `.eq-bars.paused` rule, plus a duplicate `.eq-bar` block in `theme.css` whose `wave` keyframe was being shadowed by `components.css` and had no other consumers.

No behaviour change beyond removing the animation: the row is still visibly the active one and the icon still marks the playing track.

## Test plan

- [x] Open AlbumDetail with a track playing — confirm CPU is normal (not pegged), the active row has the accent background, and the AudioLines icon is shown in the track number column instead of the bars.
- [x] Pause the track — accent background and icon stay (icon position is purely `isPlaying`-gated, so it disappears on pause as before).
- [x] Same on ArtistDetail, PlaylistDetail, Favorites, RandomMix — active row + icon look right.
- [x] Hover over the active row — the play-button still replaces the icon as before.
- [x] On Linux + NVIDIA proprietary, watch `btop` while a long album is playing on AlbumDetail — the WebProcess should sit at single-digit CPU, not ~80 %.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The **Playlists** icon in the collapsed sidebar was **off-centre** and had a **wider hover background** than every other item, because it still rendered through the expanded-mode wrapper (with `padding-right` and a `flex: 1` main link to fit the expand-toggle). Collapsed mode now reuses the **standard nav-link path** — same hitbox, same alignment as Artists, Albums, Favorites, etc.
 
+### Tracklist — drop now-playing pulse + EQ-bar animations
+
+**By [@Psychotoxical](https://github.com/Psychotoxical), PR [#488](https://github.com/Psychotoxical/psysonic/pull/488)**
+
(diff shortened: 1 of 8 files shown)


cucadmuh (approved):
LGTM

PR #523, opened 2026-05-09
refactor(ui): reorganize QueuePanel toolbar and remove redundant playback controls

kveld9:
## Summary
Refactors the Queue Panel toolbar for better visual hierarchy and removes redundant controls.
 
### QueuePanel (`src/components/QueuePanel.tsx`)
 
**Removed:**
- Crossfade button and popover (redundant — already available in Settings)
- Gapless playback button (redundant — already available in Settings)
 
**Reorganized toolbar layout:**
```
[Save] [Load] [Share] | [Infinite] [Shuffle] [Clear]
       Group 1        |       Group 2
```
 
**New grouping logic:**
- **Left group (Playlist Management):** Save playlist, Load playlist, Share queue
- **Separator:** Visual divider between management and operations
- **Right group (Queue Operations):** Infinite queue (mode toggle), Shuffle (action), Clear (destructive action)
 
## Rationale
 
1. **Reduced redundancy:** Crossfade and gapless are configuration settings that don't need quick toggling from the queue panel. They're better suited for the Settings page.
 
2. **Logical grouping:** The toolbar now separates playlist management (persistent actions) from queue manipulation (runtime actions).
 
3. **Better UX flow:** Infinite queue → Shuffle → Clear follows a natural progression from "mode setting" to "immediate action" to "destructive action".
 
## Testing
- [ ] Verify all toolbar buttons still function correctly
- [ ] Verify tooltips display properly
- [ ] Check responsive layout on smaller screens
- [ ] Ensure keyboard navigation works across groups
 
 
 ## Screenshoots
 
 ### Before
 
{D1EA6FE5-0DC3-499F-A79A-48C112139CA3}

### After

{8FE6DCFA-90DC-4A81-905C-5AB4C261C901}


cucadmuh:
It seems questionable to me to remove that functionality from quick access. I suggest we open it up for discussion.

Psychotoxical:
First of all, thanks again for your contribution. That said, I have to be honest: I’m not a big fan of this change this time. The reason is that it feels too final to me. I like the idea of grouping things by theme, but I personally use the quick buttons for Crossfade and Gapless several times a day, so I wouldn’t want them gone from there for good. What I’d be fine with is making the quick bar customisable, so everyone can choose which icons they want to see there.

kveld9:
> First of all, thanks again for your contribution. That said, I have to be honest: I’m not a big fan of this change this time. The reason is that it feels too final to me. I like the idea of grouping things by theme, but I personally use the quick buttons for Crossfade and Gapless several times a day, so I wouldn’t want them gone from there for good. What I’d be fine with is making the quick bar customisable, so everyone can choose which icons they want to see there.

Okay, the idea of ​​it being customizable is good, it hadn't occurred to me. I originally came up with this change because gappless and crossfade are typically settings on a music player that you set and forget. I don't think there's anyone who changes these settings every 2 or 3 songs, so it seemed redundant to have it visible all the time. But I repeat, the idea of ​​making it customizable is very good, I think we should work on it 👌.

PR #625, opened 2026-05-13
feat(queue): persist header duration mode + rehydrate tests

kveld9:
## Summary

- **Persist** the queue panel header duration display mode (`total` | `remaining` | `eta`) in the existing Zustand `authStore` (`localStorage` key `psysonic-auth`), instead of ephemeral `useState` in `QueuePanel`.
- **Sanitize** invalid or missing values on rehydration (same pattern as `seekbarStyle`).
- **Document** the latent risk if `partialize` is ever refactored from denylist to allowlist.
- **Tests**: trivial setter coverage + focused `computeAuthStoreRehydration` tests for the new field.

## Problem

Clicking the duration chip in the queue header cycled between total duration, time remaining, and clock ETA, but the choice lived only in component state. After a full app restart it always fell back to the default (`total`), which felt like a bug.

## Solution

- Introduce `QueueDurationDisplayMode` in `authStoreTypes` and store `queueDurationDisplayMode` with `setQueueDurationDisplayMode` (wired through `createUiAppearanceActions`, next to `setQueueNowPlayingCollapsed`).
- Default remains `'total'` for new installs and sane fallback.
- `QueuePanel` reads/writes via `useAuthStore` selectors instead of `useState`.

## Rehydration / safety

`computeAuthStoreRehydration` validates against `total` | `remaining` | `eta`. Garbage, `null`, `undefined`, or a missing key maps the patch to `'total'` so the UI never receives an unknown mode.

## Persistence note (`partialize`)

`persist` still uses a **denylist** (`musicFolders`, `musicLibraryFilterVersion` stripped); everything else is serialized. A short comment in `authStore.ts` warns that a future **allowlist** refactor must explicitly include each persisted key or values can disappear silently without a failing test unless we add round-trip coverage.

## Files touched

| File | Change |
|------|--------|
| `src/store/authStoreTypes.ts` | `QueueDurationDisplayMode` type; state + setter on `AuthState` |
| `src/store/authStore.ts` | Default `queueDurationDisplayMode: 'total'`; `partialize` comment |
| `src/store/authUiAppearanceActions.ts` | `setQueueDurationDisplayMode` |
| `src/store/authStoreRehydrate.ts` | Sanitize invalid `queueDurationDisplayMode` |
| `src/components/QueuePanel.tsx` | Replace local state with `useAuthStore` |
| `src/store/authStore.settings.test.ts` | Pin `setQueueDurationDisplayMode` |
| `src/store/authStoreRehydrate.test.ts` | **New** — corrupt / missing / valid rehydration cases |

## How to test

1. Open queue panel, click the duration chip until you land on e.g. **ETA** or **remaining**.
2. Fully quit and restart the app → same mode should still be selected.
3. (Optional) In DevTools, edit `localStorage['psysonic-auth']` and set `state.queueDurationDisplayMode` to a bogus string → reload → should reset to **`total`** without breaking the app.

```bash
npx vitest run src/store/authStore.settings.test.ts src/store/authStoreRehydrate.test.ts

### DIFF
--- a/src/components/QueuePanel.tsx
+++ b/src/components/QueuePanel.tsx
@@ -18,6 +18,7 @@ import { useCachedUrl } from './CachedImage';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 import { useAuthStore } from '../store/authStore';
+import type { QueueDurationDisplayMode } from '../store/authStoreTypes';
 import { encodeSharePayload } from '../utils/shareLink';
 import { copyTextToClipboard } from '../utils/serverMagicString';
 import { showToast } from '../utils/toast';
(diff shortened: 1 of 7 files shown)


Psychotoxical:
Thanks @kveld9! The 1.46 refactor split locales and reshuffled queue-helper exports after you opened this, so a clean merge wasn't possible. Rebuilt in #724 with you as Co-Author on the commit + credited in `settingsCredits.ts`. Closing in favor of #724.

PR #711, opened 2026-05-14
feat(ui): virtualized library card grids and configurable column cap

cucadmuh:
## Summary
- Shared virtualized card grid for library views (albums, artists, playlists, etc.) with a responsive column layout capped by user preference.
- **Settings → Appearance**: maximum library card grid columns (4–12, default **6**), persisted in auth store, with copy about performance impact.
- Branch includes a merge commit bringing **main** up to date (human duration rounding / #710).


Psychotoxical (commented):
Reviewed the shared card-grid tooling (`cardGridLayout`, `useCardGridMetrics`, `VirtualCardGrid`, `useRemeasureGridVirtualizer`) and the ~13 page conversions.

**Solid.** Per-page specifics survive the unification consistently — wrap classes, grid gaps, enter animations, item keys, selection props and row variants are all preserved; `key` correctly moved from the cards to `itemKey`; no empty/loading state or infinite-scroll sentinel lost. `libraryGridMaxColumns` wiring is complete across all six authStore files + clamp test, and rehydration-safe for installs without the field.

**Local checks:** `npm test` 877/877 · `npx tsc --noEmit` clean · `npm run build` green.

**Pushed a one-line fixup** (`c9e4c00`): the `libraryGridMaxColumns` JSDoc said "clamped 2…12", actual min is 4.

**Worth a manual pass before merge:**
- Multiple stacked virtualized grids on the shared scroll root — ArtistDetail (grouped releases) and OfflineLibrary (artist-grouped view). Multiple virtualizers against `APP_MAIN_SCROLL_VIEWPORT_ID` is a supported pattern and each grid's spacer height is stable, but it's the spot to watch for offset drift / scroll jitter.
- `VirtualCardGrid` hardcodes the scroll element to `APP_MAIN_SCROLL_VIEWPORT_ID` — an implicit contract that any consumer must live in that viewport. Fine for all current call sites.
- A few pages gained virtualization that didn't have it before (RandomAlbums etc.) — effectively a no-op at their item counts, just noting the behaviour shift.

PR #755, opened 2026-05-17
feat(playlist): virtualize tracklist + memoize rows

artplan1:
## Virtualize playlist tracklist + memoize rows

Big playlists (10k+) froze the UI — every row was mounted to the DOM.

**Change**
- Window the tracklist with `@tanstack/react-virtual` on the shared app scroll viewport (same convention as Artists / Composers / VirtualCardGrid).
- Extract a `React.memo` `PlaylistRow` with a stable callback bundle so the virtualizer's per-scroll re-renders don't cascade into every row.
- DnD preserved: drop-indicator overlay + edge auto-scroll (best-effort).

**Impact** (measured, 10,509-track playlist)
| | before | after |
|---|---|---|
| Open | ~64 s frozen | < 1 s |
| DOM rows | 10,509 | ~30 |
| Row re-renders while scrolling | ~1,600/s | ~0–10/s |


cucadmuh:
Great work — performance numbers are impressive and the `scrollMargin` approach is exactly right for a list embedded in a shared scroll viewport.

One nit worth fixing before merge:

**`getItemKey` — missing separator**

```tsx
// current
getItemKey: i => displayedSongs[i].id + i,

// suggested
getItemKey: i => `${displayedSongs[i].id}:${i}`,
```

String concatenation without a separator can produce identical keys for different `(id, index)` pairs — e.g. `id="song12"` at index `3` and `id="song1"` at index `23` both yield `"song123"`. On Navidrome (UUID-style IDs) the risk is effectively zero, but other Subsonic-compatible servers use short numeric IDs where this collision is realistic. Adding `:` costs nothing and makes the key unambiguous across all servers.

artplan1:
@cucadmuh nice catch! updated

cucadmuh:
Thanks for the PR — great work, much appreciated!

PR #756, opened 2026-05-17
fix(player): cap persisted queue window

artplan1:
## Cap persisted player queue window

Playing/shuffling a big playlist set `queue` to ~10k `Track` objects.
zustand `persist` then `JSON.stringify`'d the **whole** queue into `localStorage` on every persisted `set` → `QuotaExceededError` storm: playback never started and the main thread stalled on each write.

**Change**
`partialize` now persists only a ±250-track window around `queueIndex` (remapping the index into the slice) instead of the entire queue.

**Why it's safe**
The authoritative full queue lives **server-side** — synced via Subsonic `savePlayQueue` and restored through `getPlayQueue`. localStorage is only a fast local cache, so bounding it doesn't lose the queue; it just restores a ±250 window instantly while the full queue rehydrates from the server.

**Evidence**
Controlled before/after on a 10,509-track playlist:
- before: `QuotaExceededError` ×9, play/shuffle dead
- after: `QuotaExceeded = 0`, playback works


cucadmuh:
Thanks for the clear write-up and the controlled before/after evidence — the `QuotaExceededError` storm on big playlists was a real pain point and this is the right fix.

I pushed one extra commit with tests for the window-capping + index-remap logic (`partialize: localStorage queue window`) — five cases covering middle, near-start, near-end, small queue, and empty queue. Everything passes.

Two minor nits for whenever you get a chance (neither is a blocker):
- `const qi = state.queueIndex ?? 0` — `queueIndex` is typed `number` so the `?? 0` guard is unreachable; can just drop it.
- `PERSIST_QUEUE_HALF` would read more cleanly at module scope (mirrors `SYNC_DEBOUNCE_MS` in `queueSync.ts`).

Ready to merge from my side.

cucadmuh:
Picked up the two nits myself — pushed as a follow-up commit. Thanks for the PR!

PR #805, opened 2026-05-19
feat(favorites): virtualize songs tracklist + memoize rows

artplan1:
Mirrors the playlist virtualization fix (#755, commit `70b30f5`) for the Favorites songs list.

## Why

Opening Favorites with 10k+ starred songs mounted every row into the DOM, same condition the playlist had pre-virtualization. Subsonic `getStarred2.view` has no pagination, and Navidrome's native paginated `/api/song?starred=true` is officially undocumented/unstable — so the right fix is the proven one: window the DOM, keep the single bulk fetch.

## What

- New `FavoriteSongRow` — `React.memo` row, favorites column set (X remove, inline format, no heart/delete).
- `FavoritesSongsTracklist` now uses `@tanstack/react-virtual` with 48px estimated rows and dynamic overscan against `APP_MAIN_SCROLL_VIEWPORT_ID`.
- Stable `cb` callback bundle via `useRef` latch so the memoized row keeps referential identity across scroll-driven re-renders.
- `scrollMargin` measured with a `ResizeObserver` anchored to `.content-body` so the list re-aligns when sections above (filters, top artists, etc.) change height.
- `visibleTracks` memoized once per `visibleSongs` ref — no more `songToTrack(s)` re-mapping the full list on every row click.
- `showBitrate` plumbed from `useThemeStore` and rendered via the shared `codecLabel` helper, matching `PlaylistRow`.

## Preserved behaviors

Click/ctrl/shift select, orbit double-click, context menu (`favorite-song`), drag-out (single + bulk via `psyDrag`), preview, rate, remove, artist/album nav, header sort/resize/picker, empty state. Drag now also guards against arming on `.closest('button, a, input')` for consistency with playlist + in-file convention.

## Out of scope

No reorder / drop-indicator / edge auto-scroll (favorites has no row reorder; drag is export-only). Pagination via Navidrome native API rejected as a separate concern (undocumented, unstable).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,6 +349,14 @@ Foundational work: faster reviews, narrower diffs, and a safety net under the pa
 * Row rendering moved into a memoized `PlaylistRow` with a stable callback bundle so virtualizer scroll updates do not re-render the full list.
 * Drag-and-drop reordering is preserved: drop-indicator overlay and edge auto-scroll during drags.
 
+### Favorites — virtualized songs tracklist for large collections
+
+**By [@artplan1](https://github.com/artplan1), PR [#805](https://github.com/Psychotoxical/psysonic/pull/805)**
+
+* Opening Favorites with 10 000+ starred songs no longer mounts every row into the DOM. The songs tracklist is windowed with `@tanstack/react-virtual` on the shared app scroll viewport — same shape as the playlist virtualization fix.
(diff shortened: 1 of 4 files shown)


cucadmuh:
Thanks for #805 — favorites virtualization lands clean, same solid pattern as #755.
Join us on [Telegram](https://t.me/+GLBx1_xeH28xYTJi) or [Discord](https://discord.gg/AMnDRErm4u) if you'd like to stick around.

PR #880, opened 2026-05-27
feat(server): dual server address (LAN + public) per profile

Psychotoxical:
## Summary

- Profiles can carry an optional second address — typically a LAN counterpart of a public URL or vice versa. The connect layer probes LAN-first per profile and sticks to whichever endpoint actually answers, so the same profile gets local speed at home and the public path elsewhere without manual switching. Single-address profiles are byte-identical to before (one endpoint, one ping, same call paths).
- Same-server verification on save: both addresses are fingerprinted (ping + folders + user + license + indexes, Subsonic-generic) and compared before persist; `mismatch` / `insufficient` / `unreachable` each block save with a localized toast. Two-LAN profiles are rejected client-side; reverse case (two public) is allowed per spec.
- Share-link encoders (Orbit invites, entity / queue shares, magic-string v2) embed the **public** address by default so off-LAN guests can reach the host; a per-profile `shareUsesLocalUrl` checkbox flips that to LAN. Paste-side matchers resolve either address. Magic-string v2 carries `alt` + `shareLocal` additively; v1 invites still decode unchanged.
- Editing the primary URL to a new identifier triggers a confirm-modal-gated remigration that moves library + analysis SQLite rows (via the existing `migration_run`), the cover-cache disk bucket (new `cover_cache_rename_server_bucket` with prefer-existing merge), and every front-end keyed store (offline / hot cache / analysis + cover strategy / player queue) in one orchestrated pass. Alternate-URL-only and scheme-only edits skip the migration entirely.

## Tauri boundary

Two new additive commands — `resolve_host_addresses(hostname)` (DNS hint for the add/edit form; lookup failure swallowed to `[]` so a transient DNS hiccup never blocks save) and `cover_cache_rename_server_bucket(oldKey, newKey)` (URL-change remigration with FS-boundary sanitization rejecting absolute paths / drive letters / `..` segments). One new event — `cover:bucket-renamed` — with a listener in `useCoverArtBridge` that drops stale in-memory disk-src cache entries. No existing IPC name / arg / payload shape changed; serde ↔ TS field names verified end-to-end.

## Test plan

- [ ] Single-address regression — an existing single-URL profile connects, browses, and plays unchanged after the upgrade.
- [ ] Dual-address connect — home LAN + DDNS public on one profile. At home the badge reads "LAN" and the API/streams hit the LAN URL; on cellular the same profile flips to the public URL without a manual retry, badge updates to "extern".
- [ ] Two-LAN client-side block — saving a profile with two LAN addresses surfaces the localized "both addresses are local" toast and does not persist.
- [ ] Mismatched-server block — saving two addresses that point at different servers surfaces the "different servers" toast and does not persist.
- [ ] Orbit invite share URL — host with both addresses set: copy invite contains the public URL by default; flipping `shareUsesLocalUrl` makes the next copy carry the LAN URL.
- [ ] Magic-string v2 paste — paste a v2 invite into the AddServerForm magic field; the second address + share checkbox prefill; save persists both fields. v1 invite paste still works as a single-address save.
- [ ] URL-change remigration — edit a saved profile's primary URL to a new host: confirm modal appears with the old → new keys; on accept, library + analysis row counts stay stable (sqlite count check), cover-cache bucket has moved on disk, queue keeps playing.
- [ ] Alternate-URL-only edit no-migration — change only the `alternateUrl` (or scheme on the primary) — no confirm modal, no migration prompt.
- [ ] `./dev.sh check` — full sweep green (frontend tests + tsc + coverage + build + backend tests + clippy + llvm-cov + hot-path gates). Locally green at commit \`2851906c\`.


cucadmuh (approved):
**Approve** — rebased onto current `main` (#881–#883); conflicts resolved; frontend/tsc/vitest CI green.

Solid split of INDEX / CONNECT / SHARE, single-address backward compat (v1 magic), probe dedupe, `cover:bucket-renamed` wiring, and hot-path tests.

---

### OUT OF SCOPE (follow-up)

- **Post-save ping uses `data.url` only** (`ServersTab` add/edit) — card status can show `error` while connect would use the alternate. *Follow-up:* route through `ensureConnectUrlResolved` + `probe.baseUrl` for identity / AudioMuse.
- **`dualAddressVerifying` i18n** — keys exist, no UI yet. *Follow-up:* progress toast during verify, or drop unused keys.
- **Login + v2 magic** — persists `alternateUrl` without `verifySameServerEndpoints` (intentional: no second field). *Follow-up:* only if Login gains dual-address UI.
- **Inactive dual-address profile** — `connectBaseUrlForServer` falls back to primary until probed; cross-server queue/cover may hit the wrong endpoint. *Follow-up:* lazy probe for non-active servers.
- **`findServerIdForShareUrl`** — first match if two profiles share the same alternate. *Follow-up:* disambiguation if it shows up in practice.

---

### Manual smoke (pre-merge)

Single-address regression · LAN/public badge + endpoint flip · verify blocks (mismatch / two-LAN / unreachable off-network) · Orbit/entity share (public vs `shareUsesLocalUrl`) · magic v2 paste + save · primary URL remigration vs alternate-only / scheme-only skip.

Squash merge is fine after smoke.

PR #892, opened 2026-05-29
fix(queue): pin queueServerId on first enqueue so refs resolve

Psychotoxical:
## Summary
- Single-track enqueues from pages that don't replace the queue (Advanced Search row, SongRow + button, SongCard) left `queueServerId` null, so `seedIncoming` no-op'd, refs landed with an empty server key, and the queue panel rendered every row as the resolver placeholder (`…` / `0:00`) until a queue-replacing `playTrack` happened to pin the server.
- New `ensureQueueServerPinned` step at the entry of every add-to-queue mutation (`enqueue`, `enqueueAt`, `playNext`, `enqueueRadio`). Runs after `blockCrossServerEnqueue` so a guarded cross-server enqueue still bails without touching the pin, and after the undo snapshot so undo restores the pre-pin baseline. Idempotent: no-op when already pinned or when no active server is available.
- Regression cluster in `b1QueueServerIdentity.test.ts` covers `enqueue` / `enqueueAt` / `enqueueRadio` cache-hit after pin, the no-active-server fallback, and the already-pinned no-op.

## Test plan
- [x] `npm test` — 1435/1435 green
- [x] `npx tsc --noEmit` — clean
- [x] Targeted `npx vitest run src/store/b1QueueServerIdentity.test.ts src/store/playerStore.queue.test.ts` — 44/44 green
- [x] Manual: Advanced Search → Songs → "+" button on a row → queue panel shows the real title + duration, not "…" / "0:00"
- [x] Manual: same again after a fresh launch (no prior playback) to confirm the empty-`queueServerId` start state

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -525,6 +525,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+### Queue — new tracks no longer render as blank placeholders
+
+**By [@Psychotoxical](https://github.com/Psychotoxical), PR [#892](https://github.com/Psychotoxical/psysonic/pull/892)**
+
+* Adding tracks to the queue from Advanced Search results, song rows, or song cards right after launch could show every new entry as `…` / `0:00` instead of the real title and duration, until something else triggered a queue-replacing playback.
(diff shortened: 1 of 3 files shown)


cucadmuh:
LGTM

PR #1042, opened 2026-06-08
fix(now-playing): keep metadata cards when the track plays from local cache

Psychotoxical:
## Summary
- The Now Playing metadata fetches (album, artist info, discography, top songs) gated on `shouldAttemptSubsonicForServer(serverId, trackId)`. That guard returns false when the track's audio resolves to a `psysonic-local://` URL (hot cache / offline bytes) — correct for playback-byte calls, wrong for metadata.
- On track change the prefetched next track plays from the hot cache, so every Subsonic-backed card blanked while the Last.fm bio fallback and Bandsintown tour (non-Subsonic) stayed — and never recovered.
- Drop the trackId from the four Now Playing metadata gates so metadata loads whenever the server is reachable, regardless of where the audio bytes come from. True-offline is still covered by the online / reachability checks inside the guard.

## Scope
- Only the four Now Playing call sites change. The guard itself and every other caller (scrobble, `getSongForServer`, offline browse) are untouched; no local-library / hot-cache behaviour changes.

## Test plan
- New regression test: the metadata gate is called with the server id only, never a trackId.
- `tsc --noEmit`, `npm run build`, and `vitest` for the fetchers hook green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
+### Now Playing — cards no longer blank out on track change
+
+**By [@Psychotoxical](https://github.com/Psychotoxical), PR [#1042](https://github.com/Psychotoxical/psysonic/pull/1042)**
+
+* The "from this album", "discography" and "most played" sections on the Now Playing page disappeared after a track change once the next track started playing from the local cache, and didn't come back. The page now keeps loading that info whenever the server is reachable, regardless of whether the audio plays from cached or offline bytes.
+
 ### Servers — complete border on the active server card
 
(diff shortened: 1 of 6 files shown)


cucadmuh (approved):
LGTM — thanks for folding in the single gate and the behaviour-level regression tests from review.

**Verified (static + CI):** root cause fix, deduplicated `shouldAttemptSubsonicForServer` ownership in the fetchers hook/prewarm, guard unit test for metadata vs byte-style calls, hook test that album/discography/top songs still load when the guard would block trackId-scoped calls. CI green on the branch.

**Runtime:** I have not re-tested hot-cache track change on Now Playing yet; will spot-check when back at the desk. No blockers from my side.

**Follow-up:** filed a separate issue for the `getSongForServer` song-level metadata tail on local-cache playback (genre / play count / contributors) — out of scope for this PR, but worth a small follow-up once this lands.

PR #1043, opened 2026-06-09
fix(library-db): name slow-write ops for macOS stall diagnosis

### LINKED ISSUE #1040: [Bug]: macOS playback stall from a multi-second library-db write
Split out from #996. The device-watcher stutter in that thread (every ~3s) is a separate cause, fixed in #1039. This issue tracks a distinct stall reported by @plyr1abides on macOS ("since the 04-JUN update"): playback drops out periodically, with the audio output stalling for several seconds at a time.

### Evidence
From the #996 diagnostic build (device-watcher enumeration throttled to ~60s, so that path is ruled out), the log shows a single `library-db` write running ~50s and blocking the write connection while output stalls:

```
[library-db] SLOW write op=misc lock_wait_ms=0 exec_ms=49787
[library-db] SLOW write op=misc lock_wait_ms=49063 exec_ms=0
[psysonic] device-watcher: output stalled for 9.0s — reopening stream, pinned=None
```

The device-watcher here is only the *recovery* (it detects the stalled output and reopens the stream), not the cause. The ~50s write monopolises the single write connection and starves the render thread.

### Notes / next step
- The slow write logs as `op=misc`, the generic label shared by ~20 write call sites in `psysonic-library`, so the log doesn't say *which* write stalls. Giving these writes specific op names would let the next log pinpoint the culprit — a cheap diagnostic before any fix build.
- Likely recurring rather than a one-off sync: a ~51s poll-gap in the preceding minute points to a periodic long write.

cc @cucadmuh (psysonic-library domain)

cucadmuh:
## Summary

Diagnostic step for [#1040](https://github.com/Psychotoxical/psysonic/issues/1040) — macOS playback stalls correlated with `[library-db] SLOW write op=misc` holding the write connection for tens of seconds.

Replaces the generic `op=misc` label on **production** `LibraryStore` write paths with stable `module.action` names (`sync_state.set_sync_phase`, `track.upsert_batch_remap`, `tombstone.mark_deleted`, `cmd.purge_server`, …). Heavy paths that were already named (`bulk.finalize`, `bulk.begin`, `track.upsert_initial_ingest`, …) are unchanged.

Adds a doc comment on `with_conn` documenting the naming convention so new call sites do not regress to `"misc"`.

No behaviour change — logging/diagnostics only.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm test` — 248 files / 1903 tests + CSS import graph
- [x] `cargo test --workspace --all-targets` — all workspace crates green (`psysonic-library`: 397 passed)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] After merge: ask @plyr1abides (or local repro) to capture logs — next `SLOW write` should show a specific `op=` instead of `misc`

Fixes #1040 (diagnostic phase; follow-up fix once the culprit op is identified from logs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
+### Library DB — named slow-write ops for stall diagnosis
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#1043](https://github.com/Psychotoxical/psysonic/pull/1043)**
+
+* Production `library-db` write paths now log stable `module.action` op names instead of the generic `misc`, so the next `[library-db] SLOW write` line on macOS (or elsewhere) identifies the call site — diagnostic step for playback stalls under long write-lock holds ([#1040](https://github.com/Psychotoxical/psysonic/issues/1040)).
+
+
+
 ### Servers — complete border on the active server card
 
 **By [@Psychotoxical](https://github.com/Psychotoxical), PR [#998](https://github.com/Psychotoxical/psysonic/pull/998)**
--- a/src-tauri/crates/psysonic-library/src/browse_support.rs
+++ b/src-tauri/crates/psysonic-library/src/browse_support.rs
@@ -34,7 +34,7 @@ pub(crate) fn reconcile_album_stars(
 ) -> Result<(), String> {
     runtime
         .store
-        .with_conn("misc", |conn| {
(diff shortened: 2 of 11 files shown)


Psychotoxical (approved):
LGTM — logging/diagnostics only, no behaviour change. The `module.action` labels cover the production `LibraryStore` write paths; the remaining `misc` call sites are tests, read-only queries, or the already-named heavy paths. The next macOS `SLOW write` line should point at a concrete op for #1040.

Issue #1046, opened 2026-06-09
fix(now-playing): index-first metadata fetchers in useNowPlayingFetchers


cucadmuh:
## Context

Follow-up from PR #1042 — [fix(now-playing): keep metadata cards when the track plays from local cache](https://github.com/Psychotoxical/psysonic/pull/1042).

#1042 fixes the **metadata gate** (drop `trackId` so Subsonic section cards load while audio plays from `psysonic-local://` when the server is reachable). It does **not** change the data source: `useNowPlayingFetchers` still goes **straight to Subsonic** for every fetch.

Workdocs review: `psysonic-workdocs` → `internal/collaboration/handoffs/2026-06-09-pr-1042-now-playing-metadata-review.md`.

## Product rule (team)

**Local library index is first-class.** If SQLite has the data, read it there. Subsonic/network is **fallback only** when the index cannot serve the row (or the index is off / not ready).

Reference patterns already in-tree:

| Area | Pattern |
|---|---|
| Queue track hydration | `queueTrackResolver` — `libraryGetTracksBatch` → `trackToSong`; `getSongForServer` on miss |
| Album / artist detail | `loadAlbumFromLibraryIndex` / `loadArtistFromLibraryIndex` in `offlineLibraryIndexLoad.ts` |
| Queue panel enrichment | `useQueueTrackEnrichment` — `libraryGetTrack` + `libraryGetFacts` |

`useNowPlayingFetchers` / `prewarmNowPlayingFetchers` should join this family — **not** a fourth parallel “always network” path. Invert `offlineMediaResolve` (which is network-first → index fallback); NP loaders should be **index → network**.

## Scope — what to make index-first

| Now Playing fetch | Current | Index path (preferred) | Network fallback |
|---|---|---|---|
| **Album card** (`albumData`) | `getAlbumForServer` | `loadAlbumFromLibraryIndex(serverId, albumId)` | `getAlbumForServer` (metadata gate, no `trackId`) |
| **Discography** | `getArtistForServer` → albums | `loadArtistFromLibraryIndex` → `.albums` | `getArtistForServer` |
| **Most played** (`topSongs`) | `getTopSongsForServer` | New small helper: index tracks for artist, sort by `play_count` desc, limit 5 (advanced search or reuse lossless-browse + sort pattern from `useArtistDetailData`) | `getTopSongsForServer` |
[...]

PR #1049, opened 2026-06-09
fix(now-playing): index-first metadata fetchers (#1046)

Psychotoxical:
## Summary

Follow-up to #1042. `useNowPlayingFetchers` / `prewarmNowPlayingFetchers` now read the local library index first and fall back to Subsonic only on index miss / index off / not ready — the same index-first family as `queueTrackResolver` and `useQueueTrackEnrichment`, not a new always-network path.

- New `utils/library/nowPlayingMetadataResolve.ts`: index→network resolvers for **album, discography, top songs, song-level meta**, reused by both the hook effects and prewarm.
- On a populated, ready index those four cards render from SQLite with **no Subsonic call** (so genre / play count / contributors / cards survive hot-cache & offline-pinned playback); index miss/off/not-ready falls back to the network path unchanged.
- `artistInfo` (bio/similar) has no index source and stays network-only. The global `shouldAttemptSubsonicForServer` guard is **not** weakened — scrobble and playback-byte paths keep the `trackId` + `psysonic-local://` skip.

## Test plan

- [x] New `nowPlayingMetadataResolve.test.ts`: per fetcher index hit → no Subsonic call, index miss → network fallback, index off → network fallback (12 cases)
- [x] Existing `useNowPlayingFetchers.test.ts` id-gating + #1042 guard regressions stay green
- [x] `npx tsc --noEmit`
- [x] `./dev.sh frontend-check` (tests, tsc, coverage gate, CSS import graph)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -167,6 +167,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+### Now Playing — metadata reads from the local library index first
+
+**By [@Psychotoxical](https://github.com/Psychotoxical), PR [#1049](https://github.com/Psychotoxical/psysonic/pull/1049)**
+
+* The "from this album", "discography", "most played" and song details on the Now Playing page now come from the local library index when it has them, only falling back to the server when the index can't serve a row. Cards and fields (genre, play count, contributors) stay populated during cached and offline playback, with fewer server requests.
+
+
+
 ### Library DB — named slow-write ops for stall diagnosis
 
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1043](https://github.com/Psychotoxical/psysonic/pull/1043)**
--- a/src/hooks/useNowPlayingFetchers.ts
+++ b/src/hooks/useNowPlayingFetchers.ts
@@ -86,10 +86,9 @@ export async function prewarmNowPlayingFetchers(
   } = deps;
 
   if (!fetchEnabled || !subsonicServerId) return;
-  // The only network gate in this function. No trackId: metadata must be fetched
-  // even when the track's audio is local (hot-cache / offline). Offline is still
-  // covered by the online/reachability checks inside the guard.
-  if (!shouldAttemptSubsonicForServer(subsonicServerId)) return;
+  // Index-first resolvers run whenever there's a server id (offline included) —
+  // each guards its own network fallback. artistInfo below is the one
+  // network-only job, so it keeps the reachability gate.
 
   const jobs: Array<Promise<unknown>> = [];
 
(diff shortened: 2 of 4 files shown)


cucadmuh (dismissed):
LGTM — this matches #1046 well: shared index-first resolvers, hook + prewarm wired, `artistInfo` correctly network-only, guard untouched, 12-case resolver matrix + existing #1042 hook regressions. CI green on my read.

Two **recommended in-PR improvements** (UX, not blockers):

### 1. Let index reads run when the server is unreachable

Today `subsonicFetchAllowed` still requires `shouldAttemptSubsonicForServer(subsonicServerId)` (reachable + online) before any effect runs — so the new index arms never execute in true offline even when SQLite has the rows. That’s pre-#1049 behaviour, but it limits the main win of index-first for offline-pinned / disconnected playback.

**Suggestion:** split gates in `useNowPlayingFetchers` / prewarm:
- **Index path:** `fetchEnabled && subsonicServerId` (try resolvers whenever we have a playback server id)
- **Network path:** existing metadata gate inside each resolver’s fallback arm only (`shouldAttemptSubsonicForServer(serverId)`, no `trackId`)

Keep `artistInfo` on the network-only gate. Last.fm / Bandsintown unchanged.

### 2. Tighten local top-songs selection

`resolveNpTopSongs` FTS on `artistName` + client `artistId` filter is a reasonable v1, but name collisions can return the wrong slice before network fallback kicks in.

**Suggestion (still in-tree, no new Rust API):** prefer a more deterministic index query before FTS — e.g. pull tracks from albums already resolved for that `artistId` (discography index hit or a narrow album-id restrict), sort by `play_count`, cap at 5; keep `getTopSongsForServer` as fallback on empty/miss. Would align top songs with the same artist bucket discography already uses.

Happy to approve as-is if you prefer a follow-up issue for (1) and (2) — both are small and would make the index-first story land harder for offline users.

Psychotoxical:
Both in-PR points addressed in `ab083e3d`:

1. **Index reads run offline** — split the hook/prewarm gate: index resolvers run on `fetchEnabled && serverId`, and the reachability guard moved into each resolver's network fallback arm. `artistInfo` (no index source) stays network-only.
2. **Deterministic top songs** — `resolveNpTopSongs` now derives from the artist's own discography albums sorted by `play_count` (same bucket the discography card uses) instead of an FTS-on-name query, so name collisions can't surface the wrong tracks. `getTopSongsForServer` remains the fallback on miss.

Resolver tests extended with the unreachable (index-only) cases; tsc + `./dev.sh frontend-check` green.

cucadmuh (approved):
Re-review after `ab083e3` — both suggested in-PR improvements are in and look correct. **Approve.**

### 1. Offline index reads ✅
- Hook/prewarm: `indexFetchAllowed = fetchEnabled && serverId` for the four `resolveNp*` paths; `networkOnlyAllowed` only for `artistInfo`.
- Resolvers: `shouldAttemptSubsonicForServer(serverId)` moved into each network fallback arm; index arm runs when unreachable.
- Tests: unreachable + index miss → no Subsonic call (`getAlbumForServer` / discography / top songs).

### 2. Deterministic top songs ✅
- `resolveNpTopSongs` walks the artist's discography albums → `libraryGetTracksByAlbum` per album → sort by `play_count` desc → cap 5.
- No FTS-on-name; same artist bucket as discography. Network fallback unchanged on miss/off/unreachable.
- Test asserts `t-hi` before `t-lo` by play count.

CI green on the branch. Closes #1046 from my side — nice follow-up to #1042.

PR #1063, opened 2026-06-11
fix(favorites): reflect player-bar song star in track lists

artplan1:
## What

Liking a song from the **player bar** (also fullscreen / global shortcuts) now stays starred in the track lists — album detail, Favorites, Random Mix, playlists. Previously the like appeared not to apply to the list, even though toggling from a list row correctly updated the player bar.

## Root cause

All star toggles go through `queueSongStar`, which sets a session `starredOverrides[id]` optimistically and then syncs to the server. On success, `onStarSuccess` **deleted** that override and patched only `currentTrack` + the queue resolver cache.

List views seed their starred state from a one-shot fetch and reflect later changes **only** by merging `starredOverrides`. So the two directions were asymmetric:

- list row → like: updates the row's own local state **and** the override → both stick ✅
- player bar → like: sets **only** the override → list shows it briefly → sync deletes the override → row reverts to its stale fetched value ❌

Online the sync resolves near-instantly, so the list appears to "not update" at all.

## Fix

Keep the star override as the durable session source of truth — stop deleting it on success. It is unpersisted (not in the player-store `partialize`, reset on reload) and superseded by the next toggle, so it never diverges from the server. The merge logic in every list view already handles persistent `true`/`false` overrides. The `currentTrack` / queue-cache patches stay so the persisted current track is honest across reload. Ratings keep their existing clear-on-success behavior.

## Testing

- `vitest run src/store/pendingStarSync.test.ts` — assertions updated (override kept, not cleared) and a redundant test dropped; 5 pass.
- Manually verified in `tauri:dev`: star from the player bar now toggles and persists in the track list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,6 +349,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+### Favorites — player-bar star stays synced in track lists
+
+**By [@artplan1](https://github.com/artplan1), PR [#1063](https://github.com/Psychotoxical/psysonic/pull/1063)**
+
+* Liking a song from the **player bar**, fullscreen player, or global shortcuts now updates the star in album tracklists, playlists, Random Mix, and Favorites — the row no longer reverts the instant the server sync completes.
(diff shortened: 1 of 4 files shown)


cucadmuh (dismissed):
LGTM

cucadmuh (approved):

PR #1079, opened 2026-06-13
fix(installer): stderr logging, /dev/tty reinstall prompt, hardened download

### LINKED ISSUE #1078: [Bug]: install.sh: [INFO] log lines get captured into the download URL — curl fails with "bad range in URL position 4"
### What's wrong?

The curl | sudo bash installer (scripts/install.sh) fails on apt-based systems before any download happens: the [INFO] log lines printed inside get_download_url() are captured into the download URL variable via command substitution, and curl rejects the mangled multi-line "URL" with `bad range in URL position 4`.

### Steps to reproduce

1. On any Debian/Ubuntu (apt-based) system, run: curl -fsSL https://raw.githubusercontent.com/Psychotoxical/psysonic/main/scripts/install.sh | sudo bash
2. Script detects apt and resolves the latest release (app-v1.47.0) successfully.
3. The download step fails immediately with curl exit code 3, and set -e aborts the script.

### What should happen?

The latest .deb (Psysonic_1.47.0_amd64.deb) is downloaded and installed via apt.

### What actually happens?

No download is attempted. curl exits with `curl: (3) bad range in URL position 4:` and echoes back the mangled "URL" — two [INFO] log lines followed by the real URL. Full terminal output in Logs below. Manually downloading the same asset URL and running `sudo apt install ./Psysonic_1.47.0_amd64.deb` works fine, so the release asset itself is good.

### Psysonic version

1.47.0 (version resolved by the installer — app never installed, so no Settings → About to read)

### Operating system

Linux

### Install source

Other (please describe in "Anything else")

None

### Server version

_No response_

### Logs

```shell
kb@G434:~/Desktop/concord$ curl -fsSL https://raw.githubusercontent.com/Psychotoxical/psysonic/main/scripts/install.sh | sudo bash

╔══════════════════════════════════════════════════════════╗
║           Psysonic Auto-Installer                        ║
║   Install the latest release from GitHub Releases        ║
╚══════════════════════════════════════════════════════════╝

[INFO] Detected Debian/Ubuntu-based system (apt)
[INFO] Download URL: [INFO] Fetching latest release information...
[INFO] Latest version: app-v1.47.0
https://github.com/Psychotoxical/psysonic/releases/download/app-v1.47.0/Psysonic_1.47.0_amd64.deb
[INFO] Downloading package...
curl: (3) bad range in URL position 4:
[INFO] Fetching latest release information...
[INFO] Latest version: app-v1.47.0
https://github.com/Psychotoxical/psysonic/releases/download/app-v1.47.0/Psysonic_1.47.0_amd64.deb
   ^
```

### Screenshots / video

_No response_

### Anything else

**Root cause** (from reading `scripts/install.sh` on `main`):

`get_download_url()` returns its value by echoing to stdout, and `main()` captures it with command substitution (`download_url=$(get_download_url)`). But the `info()` helper *also* writes to stdout, so the substitution captures the two `[INFO]` lines — ANSI color escapes included — plus the URL. `$download_url` ends up a three-line string. When that reaches `curl -L -o "$package_file" "$download_url"`, curl's URL globbing treats the `[` from the escape sequence as the start of a character range and aborts with `bad range in URL position 4` before making any request; `set -e` then kills the script.

**Suggested fix** — route log helpers to stderr so command substitution only captures real return values:

```bash
info() {
    echo -e "${BLUE}[INFO]${NC} $1" >&2
}
```

(same for `warn()`, and optionally `success()`/`error()` for consistency). Adding `--globoff` and `--fail` to the download curl call would be cheap defense in depth.

Side note spotted in the same file: `check_installed()` uses `read -p`, which misbehaves under the documented `curl … | sudo bash` invocation — `read` consumes the next line of the script from the pipe instead of reading the keyboard. Only triggers on the reinstall path, but worth fixing in the same pass (e.g. `read … < /dev/tty`).

Environment: Ubuntu, x86_64, installed via the `curl | sudo bash` one-liner.

Happy to open a PR for this if useful.

kbennett2000:
Fixes #1078

## What changed
`scripts/install.sh` only, four commits:

1. The four log helpers (`info`, `success`, `warn`, `error`) now write to **stderr**, so functions returning values over stdout (`get_download_url`) stay clean inside command substitution.
2. The reinstall prompt in `check_installed()` reads from `/dev/tty` instead of stdin.
3. The two package-download `curl` calls (apt and dnf/yum branches) gain `--fail --globoff`.
4. The reinstall prompt degrades gracefully when no controlling terminal can be opened: warn and `exit 0` instead of failing under `set -e` (review suggestion).

## Why
- **stderr logging:** `main()` captures `get_download_url()` with command substitution; with `info()` on stdout, `$download_url` held two `[INFO]` lines (ANSI escapes included) plus the URL. curl's URL globber trips on the `[` characters and aborts with `curl: (3) bad range in URL position 4` before any request; `set -e` then kills the install. Side benefits: `error()` messages raised inside substituted functions are now visible instead of being swallowed, and the `[INFO]` lines print in their natural order.
- **/dev/tty read:** under the documented `curl … | sudo bash` invocation, stdin is the script stream, so a bare `read` consumes the next line of the script as the user's "answer" rather than keyboard input.
- **--fail --globoff:** on an HTTP error the old call saved the error page as the `.deb`/`.rpm` and handed it to the package manager; `--fail` aborts instead. `--globoff` makes curl treat the URL literally as defense in depth against bracket characters.
- **terminal probe:** the guard opens `/dev/tty` rather than testing `[ -r /dev/tty ]`, because the node is present and mode 0666 even without a controlling terminal — `test -r` passes there while the subsequent open still fails with ENXIO. Probing with an actual open catches both the no-ctty case and a missing node.

## Who should notice
End users installing with the `curl | sudo bash` one-liner. The dnf/yum branch shares the same helpers and download call, so it gets the same fixes. No app code touched — developers and existing installs unaffected.

## How to verify
- `bash -n scripts/install.sh`; shellcheck (0.9.0) reports zero findings before and after.
- On an apt-based machine: `sudo bash scripts/install.sh` from this branch resolves the latest release, downloads, and installs.
- URL-capture fix: ran the script before/after against a stub `curl` (fixture for the GitHub API call; real curl binary doing URL *parsing* for the download call) and a stub `apt`, as root on Ubuntu. Before: download curl invoked with the mangled three-line URL, exits 3, `apt` never reached. After: clean single-line URL, download proceeds, `apt install -y …` invoked.
- /dev/tty fix: allocated a pty with `script(1)`, put a fake `psysonic` on `PATH` to trigger the reinstall prompt, then fed *different* answers to stdin (pipe) and the terminal. Old behavior follows the pipe in both directions; fixed behavior follows the terminal in both directions.
- Terminal probe: under `setsid` (no controlling terminal) with the package "already installed", the script now warns and exits 0; the same scenario with a bare `[ -r /dev/tty ]` guard passes the test but still dies on the read with ENXIO. Interactive pty behavior is unchanged in both the y and n directions, and the fresh-install path never reaches the prompt.

## Scope notes
- Tauri boundary not touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -396,6 +396,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 
 
+### Auto-install script — `curl | sudo bash` works again
+
+**By [@kbennett2000](https://github.com/kbennett2000), PR [#1079](https://github.com/Psychotoxical/psysonic/pull/1079)**
+
+* The Linux auto-installer (`install.sh`) failed before any download because `[INFO]` log lines were captured into the package URL and curl rejected the mangled string; logging now goes to stderr, the reinstall prompt reads from the terminal, and package downloads use `--fail --globoff`.
+
+
+
 ## [1.47.0]
 
 > **🙏 Thank you to our amazing Discord community.** This release would not have been possible without your tireless support, quality checks, bug reports and all-round collaboration. Every report, every repro and every bit of feedback shaped what shipped here — thank you. Come join us: [discord.gg/AMnDRErm4u](https://discord.gg/AMnDRErm4u)
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -107,7 +107,7 @@ install_package() {
     
     if [ "$OS_TYPE" = "debian" ]; then
         package_file="${package_file}.deb"
-        curl -L -o "$package_file" "$download_url"
+        curl --fail --globoff -L -o "$package_file" "$download_url"
         
         info "Installing package..."
         $PACKAGE_MANAGER install -y "$package_file" || {
@@ -116,7 +116,7 @@ install_package() {
         }
     elif [ "$OS_TYPE" = "rhel" ]; then
         package_file="${package_file}.rpm"
-        curl -L -o "$package_file" "$download_url"
+        curl --fail --globoff -L -o "$package_file" "$download_url"
         
         info "Installing package..."
         $PACKAGE_MANAGER install -y "$package_file"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -130,7 +130,9 @@ install_package() {
 check_installed() {
     if command -v $APP_NAME &> /dev/null || command -v ${APP_NAME^} &> /dev/null; then
         warn "${APP_NAME} appears to be already installed."
-        read -p "Do you want to reinstall? (y/N): " -n 1 -r
+        # Under `curl ... | bash`, stdin is the script stream itself, so
+        # read the answer from the controlling terminal instead.
+        read -p "Do you want to reinstall? (y/N): " -n 1 -r < /dev/tty
         echo
         if [[ ! $REPLY =~ ^[Yy]$ ]]; then
             info "Installation cancelled."
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -131,9 +131,16 @@ check_installed() {
     if command -v $APP_NAME &> /dev/null || command -v ${APP_NAME^} &> /dev/null; then
         warn "${APP_NAME} appears to be already installed."
         # Under `curl ... | bash`, stdin is the script stream itself, so
-        # read the answer from the controlling terminal instead.
-        read -p "Do you want to reinstall? (y/N): " -n 1 -r < /dev/tty
-        echo
+        # read the answer from the controlling terminal instead. Probe by
+        # opening: `[ -r /dev/tty ]` passes on the 0666 device node even
+        # without a controlling terminal; only open() reports the failure.
+        if { : < /dev/tty; } 2>/dev/null; then
+            read -p "Do you want to reinstall? (y/N): " -n 1 -r < /dev/tty
+            echo
+        else
+            warn "No terminal available for prompt; skipping reinstall."
+            exit 0
+        fi
         if [[ ! $REPLY =~ ^[Yy]$ ]]; then
             info "Installation cancelled."
             exit 0
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -14,20 +14,22 @@ YELLOW='\033[1;33m'
 BLUE='\033[0;34m'
 NC='\033[0m' # No Color
 
+# Log helpers write to stderr so functions that return values via stdout
+# (e.g. get_download_url) stay clean when called in command substitution.
 info() {
-    echo -e "${BLUE}[INFO]${NC} $1"
+    echo -e "${BLUE}[INFO]${NC} $1" >&2
 }
 
 success() {
-    echo -e "${GREEN}[SUCCESS]${NC} $1"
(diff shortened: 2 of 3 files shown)


kbennett2000:
Pushed the /dev/tty read and the --fail/--globoff hardening as discussed — PR now covers everything from #1078.

cucadmuh:
LGTM — this now covers all of #1078: stderr logging fixes the URL pollution, the `/dev/tty` read fixes the reinstall prompt under `curl | sudo bash`, and `--fail --globoff` is sensible hardening. Minimal, well-scoped, one commit per fix. Verified the diff and the reasoning holds for both the apt and dnf/yum branches.

One optional recommendation worth folding in: guard the `/dev/tty` read for the no-controlling-terminal case so it degrades gracefully instead of failing under `set -e`, e.g.

```bash
if [ -r /dev/tty ]; then
    read -p "Do you want to reinstall? (y/N): " -n 1 -r < /dev/tty
    echo
else
    warn "No terminal available for prompt; skipping reinstall."
    exit 0
fi
```

Not a blocker — happy to merge as-is and treat this as a tiny follow-up.

kbennett2000:
Folded in as a fourth commit — small enough to do now rather than as a follow-up.

One deliberate deviation from the snippet: the guard probes by actually opening `/dev/tty` instead of testing `[ -r /dev/tty ]`. On Linux the node is present and mode 0666 even when the process has no controlling terminal, so `test -r` (an access(2) permission check) passes — and the `read < /dev/tty` then dies under `set -e` anyway, which is exactly the failure the guard is meant to prevent. Only an actual open() reports it:

```
$ setsid -w bash -c '[ -r /dev/tty ]; echo "test -r /dev/tty: exit $?"; : < /dev/tty; echo "open /dev/tty: exit $?"'
test -r /dev/tty: exit 0
bash: line 1: /dev/tty: No such device or address
open /dev/tty: exit 1
```

So the condition is:

```bash
if { : < /dev/tty; } 2>/dev/null; then
    read -p "Do you want to reinstall? (y/N): " -n 1 -r < /dev/tty
    echo
else
    warn "No terminal available for prompt; skipping reinstall."
    exit 0
fi
```

Everything else (warn text, `exit 0`) matches your suggestion. Verified both ways: headless under `setsid` with the app "already installed" now warns and exits 0, where the literal `-r` form passes its guard and still errors on the read; interactive pty runs are unchanged in both the y and n directions. Happy to swap to the literal `-r` form if you'd rather keep it simpler.

cucadmuh:
The open-probe ({ : < /dev/tty; } 2>/dev/null) is the right call — sharper than my [ -r /dev/tty ] suggestion, since the 0666 node makes access(2)/test -r pass without a controlling terminal and the read would still die on ENXIO. Your version catches both the no-ctty and missing-node cases. Verified the final diff: interactive y/N path unchanged, headless now warns and exits 0, and all of #1078 (stderr logging, /dev/tty read, --fail --globoff) is covered. LGTM — good to merge.

PR #1185, opened 2026-06-25
feat(i18n): add Polish (pl) translation

Rextens:
Added polish locale, nothing else. 

Few things are missing bcs of new commits, and probably commits names are not in the psysonic's convention. Please, write delta and whether I should change the commit names.


cucadmuh (approved):
lgtm

PR #1186, opened 2026-06-25
feat(album): show all genres in album details, linkable to genre pages

### LINKED ISSUE #1184: [Feature]: Show multiple genre in the album details, and make them clickable
### What's the use case?

Semi-related to #1067

The album details only shows a single genre. The genre tag area does properly list an album in multiple genres.

### Proposed solution

When looking at album details, all genres should be shown.

Further, since this feature request would be work done in the album details, it would be nice to have the genre clickable to instantly navigate to the genre.

### Alternatives considered

_No response_

### Anything else

_No response_

Psychotoxical:
Closes #1184.

Album detail showed only a single genre. It now surfaces every genre a release spans and makes them navigable.

- Genres are the union of album tags and track tags (case-insensitive dedupe, album order first), matching what genre browse derives — via a new `deriveAlbumGenreTags` helper.
- The main genre shows inline; a `+N` chip opens a cursor-anchored menu (context-menu style) with the remaining genres, each linking to its genre page.
- Album detail now reads from the local SQLite index when it is ready (not only with favorites-offline), so genres match genre browse and the page works offline — `resolveAlbum` is index-first with the network as fallback.
- `genreColor` extracted to a shared util (reused by the Genres page); new i18n keys across all 11 locales.

Test plan: `deriveAlbumGenreTags` unit tests + `AlbumHeader` tests (cursor menu, track union, single genre, legacy split); `./dev.sh frontend-ci` green; `tsc` + ESLint clean.

No Tauri boundary changes.


cucadmuh (approved):
lgtm

PR #1241, opened 2026-07-05
feat(library): multi-library filter — browse & search across selected libraries

### LINKED ISSUE #1220: [Feature]: Enable multiple libraries
### What's the use case?

In Navidrome web UI and most clients, it's possible to select/enable multiple libraries on a server, Psysonic currently only supports selecting a single library or "All libraries".

### Proposed solution

Allow enabling multiple different libraries by adjusting the main top-left Library drop-down logic:
- Clicking an item in the drop-down should toggle it on/off (using existing checkmark for visual feedback)
- Drop-down no longer closes on clicking an item, to allow toggling multiple different libraries while open
- Only clicking outside the drop-down or clicking the main Library button that opens the drop-down will close the drop-down

Available content would become the aggregate of whatever libraries are enabled.

### Alternatives considered

_No response_

### Anything else

_No response_

cucadmuh:
Closes #1220.

## Multi-library filter — browse & search across selected libraries

Lets you pick **several music libraries at once** (Subsonic "music folders" / Navidrome libraries) and have every browse, search, genre and detail view show their **combined** content, with the same album/artist/track that lives in more than one selected library collapsed into a **single aggregated entry**. The whole thing is built **over the local SQLite index** and is **speed-first**: it stays in the tens-of-milliseconds range on a real 212k-track database.

This is a large branch (one squashable feature, ~40 commits) because the naive version of this feature is a performance trap. The bulk of the work is making the combined view *fast* on big libraries, and shaping the data model so the same mechanism extends to cross-server mixing later without a rewrite.

---

## Why

Two independent reasons:

1. **Generic Subsonic can't express it.** Navidrome can filter by multiple libraries server-side (repeated `musicFolderId` → `library_id IN (...)`), but generic Subsonic clients can only send a single `musicFolderId`. So a multi-folder combination has to be done **client-side**, and the only place we can do it cheaply and offline is the **local index** that already mirrors the whole server (rule P20).

2. **The previous attempt was slow.** An earlier "combine libraries" try degraded badly on large libraries. Before writing any code this time, I profiled a real 668 MB / 212,494-track DB (SQLite 3.51, warm cache) and pinned down exactly *why* it was slow, then designed around it. The design doc with the raw timings lives in the team workdocs (`2026-07-multi-library-filter`).

The old filter was **single-select** only: `musicLibraryFilterByServer: Record<string,'all'|string>` — one library or all, no merge.

---

## What was slow, and the rules that come out of it

Five measured root causes (timings on a real large test DB, ~212k tracks):

- **A — non-sargable scope filter.** The scope helper was `COALESCE(NULLIF(library_id,''), json_extract(raw_json,'$.libraryId'), …) = ?`. That expression can't use the `library_id` index, so it falls back to an index scan **plus a JSON parse per row** — scoped album `GROUP BY` was **370–400 ms**. A sargable `library_id IN ('1','2')` hits `COVERING INDEX idx_track_library` and is effectively instant.
- **B — naive FTS join order (the big one).** `track_fts f JOIN track t ON t.rowid=f.rowid WHERE MATCH … ORDER BY bm25` lets the planner drive from `track` and re-run the MATCH per row → **13–14 s**. FTS-first `EXISTS` (FTS stays the driving table, the row predicates are a correlated existence check) → **6–11 ms**. A ~2000× difference.
- **C — query-time normalization/dedup.** Normalizing/hashing artist/album strings on every browse to merge duplicates is O(rows) string work per query. Keys must be **precomputed and indexed**.
- **D — missing `ANALYZE`.** `sqlite_stat1` had never been populated, so the planner mis-estimated FTS selectivity and picked bad join orders. Cheap to fix.
- **E — off-index `library_id` on aggregations.** Scoped album `GROUP BY` was 145 ms vs 18 ms unscoped; a covering `(server_id, library_id, album_id)` index brings scoped down to ~13 ms.

Normative query rules that the whole branch follows: **sargable `library_id IN (…)`**, **FTS-first `EXISTS` everywhere**, **precomputed identity keys**, **new covering indexes**, **`ANALYZE`**.

---

### Ordered scope model

The selection is an **ordered list of `(server_id, library_id)` pairs**, index 0 = highest priority. Order decides the dedup winner ("first library in the list wins"). The exact same structure serves within-server (one server, several libraries — what ships here) and cross-server (several servers — later): only the number of pairs grows. Persistence is **additive** — the legacy `musicLibraryFilterByServer` (`'all'|string`) stays for back-compat, and a new ordered `musicLibrarySelectionByServer` is written alongside it (legacy field is derived from the selection).

### Two-layer query model

**Layer 1 — per-`(server, library)` selection (always).** Predicate is `server_id = ? AND deleted = 0 AND library_id IN (?, …)`, sargable and covered by the new indexes. FTS reads are FTS-first `EXISTS`. When the whole server is selected ("all"), the `IN` clause is omitted so the existing `(server_id, …)` indexes are used.

**Layer 2 — identity dedup + priority (only when >1 pair is selected).** Rows are collapsed by a **precomputed identity key** and the row with `MIN(priority_rank)` among the selected pairs wins. A single-library selection skips this entirely — no dedup overhead. Rows with an **empty key** (missing tags) never merge and are shown individually.

- `album_key` = `norm(album_artist|artist) + norm(album)`
- `artist_key` = `norm(artist)`
- `cluster_key` = `norm(artist|title|album)` (tracks/search), with a ±5 s duration guard to split radio-edits / bad tags

**Aggregated detail.** Album/artist detail is a **virtual aggregate**: the incoming per-server entity id resolves to its `album_key`/`artist_key`, tracks are unioned across the selected libraries and deduped by `cluster_key` + priority, and album/artist-level metadata (cover, year, bio) comes from the **highest-priority owner that actually has it**, with per-field fallback down the order.

### Where the identity keys live: `library-cluster.db`

Keys are stored **denormalized** (with `server_id` + `library_id`) in a **separate, `ATTACH`-ed, rebuildable** database, reusing the existing `library-cluster.db` rather than adding a new file. Denormalizing lets a single covering-index scan do scoped dedup (**16 ms** vs **52 ms** for a cross-attach JOIN back to `track`), and it costs nothing extra on the sync path because the table is **fully rebuilt in a batch pass** from `track`.

```sql
CREATE TABLE track_cluster_key (
  server_id TEXT NOT NULL, library_id TEXT NOT NULL, track_id TEXT NOT NULL,
  cluster_key TEXT, album_key TEXT, artist_key TEXT, duration_sec INTEGER,
  PRIMARY KEY (server_id, track_id)
);
CREATE INDEX idx_ck_scope_album  ON track_cluster_key(server_id, library_id, album_key);
CREATE INDEX idx_ck_scope_artist ON track_cluster_key(server_id, library_id, artist_key);
CREATE INDEX idx_ck_scope_track  ON track_cluster_key(server_id, library_id, cluster_key);
CREATE TABLE cluster_meta (key TEXT PRIMARY KEY, value TEXT);  -- norm_version, build_at
```

It never touches the library schema, is excluded from library backup/restore, and is always rebuildable — a corrupt/migrated identity file can never endanger the library. It is attached (read-only on the read connection) and created-if-missing on both store connections; the file-swap/restore path deliberately skips its sidecars.

---

## Backend changes (Rust — `psysonic-library`)

- **Migration 016 (`multi_library_scope`)** — covering indexes on `track` for sargable scoped browse (`(server_id, library_id, album_id)`, `(server_id, library_id, artist_id)`, title/genre variants), plus `ANALYZE`.
- **Migration 017 (`library_tag_state`)** — per-server library-tagging state so the post-sync tagging pass is resumable.
- **`library_id` backfill (one-time).** `repair_library_id_from_raw_json` projects `library_id` from `raw_json` (`$.libraryId` / `$.library_id` / `$.musicFolderId`) where the hot column is empty, gated by a `library_data_migration` marker so it runs once, followed by `ANALYZE`. After this the hot `library_id` column is authoritative and the `json_extract` fallback leaves the hot path. Servers that never expose per-track library stay `library_id=''` → treated as a single implicit library, visible only under "all".
- **`identity/` module** — `norm` (string normalization), `keys` (build `cluster_key`/`album_key`/`artist_key`), `attach` (attach/create `library-cluster.db`), `rebuild` (batch rebuild + `norm_version` gate).
- **`scope_merge.rs`** — the Layer-2 priority-deduped, aggregated reads over the ordered scope (single covering-index scan, priority via `MIN(rank)`).
- **Search paths converted to the fast pattern** — `search.rs` (sargable helpers + FTS-first `EXISTS`), `advanced_search.rs`, `live_search.rs`, and genre/album browse all use `library_id IN (…)` and FTS-first `EXISTS`, and thread the ordered multi-scope through when >1 pair is selected.
- **Post-sync integration** — a per-folder library **tagging pass** runs after whole-server bulk syncs and delta syncs, and a **cluster rebuild** runs after a member's sync so keys stay current; resync no longer wipes an already-populated `track.library_id`. Live search shows an "index incomplete" banner while keys are still building.

### Normalization is locale-aware, and gated

`norm(s)` = Unicode NFD → drop combining marks → lowercase → keep letters/digits only; an empty result makes the key `NULL` so that row never merges. The diacritic/ligature folding table is kept in step with the **UI locales we actually ship**: German `ß→ss`, Norwegian/Danish `æ→ae`, French `œ→oe`, Romanian comma-below `ș/ț`, and Cyrillic `ё→е` / `й→и`. **CJK is intentionally left verbatim** — Han/Kana are kept as-is and Japanese dakuten are phonemic (`が` must not fold to `か`).

Normalization is versioned by `cluster_meta.norm_version`; this branch sets it to `2`. The read path (`ensure_cluster_keys_built`) now honors that gate: a `norm_version` mismatch forces a rebuild of **all** servers (a single global version stamp means a per-server rebuild would strand other servers' stale keys). `CONTRIBUTING.md` documents "extend the normalizer when you add a locale, then bump `NORM_VERSION`" as an explicit step.

---

## Frontend changes (`src/`)

- **Ordered selection state.** `musicLibrarySelectionByServer` (ordered) added alongside the legacy filter; `setMusicLibrarySelection` keeps the legacy `musicLibraryFilterByServer` in sync via a derived value so every existing reader keeps working. `musicLibraryFilterVersion` bumps drive the ~30 hooks/pages that refetch on scope change.
- **Multi-select sidebar picker** with priority ordering, restoring the server-cluster library visualization.
- **Scope threaded through** the local browse/search wrappers and API layer; single-scope behavior is byte-for-byte identical when the multi field is absent (additive Tauri contract, rule 04).
- **Perf polish on browse:** skeleton-first paint, catalog-cache hydration in `useLayoutEffect` before paint, deferred background SQL, deferred catalog reload after picker selection (via `requestAnimationFrame` coalescing), route-chunk prefetch, and PsyLab-gated browse traces for All Albums/Artists.
- **Architecture/layering.** To keep `src/lib` at the dependency floor and the import graph acyclic (CI `dep:check`), the store→lib catalog-reload side effect is inverted through a registration seam (`musicLibraryFilterNotify` + an app-layer bridge registered from `AppShell`), the debug-logging gate is injected into a lib-safe module instead of importing the store, and the browse route-prefetch/nav helpers moved from `src/lib/library` into the album/artist/sidebar feature folders.

---

## Backward compatibility & limitations

- **Additive persistence** — old `musicLibraryFilterByServer` is preserved; upgrading users keep their single-library choice (migrated to a one-element ordered selection; `'all'` → all libraries of that server).
- **Additive Tauri boundary** — existing commands keep their single `library_scope: Option<String>`; multi-scope is an optional extra input.
- **Generic Subsonic** servers with no per-track library stay a single implicit library and only appear under "all"; multi-library dedup applies only where libraries are actually distinguishable.
- **No `album` table for v1** — album browse is `track GROUP BY` + the new covering index (13–18 ms, 16 ms with key-dedup); materializing an album-per-`album_key` table only matters at millions of tracks.
- **Cross-server mixing is out of scope for the shipped UI** but the data model is shaped for it: identity keys are global per track and the scope is a list, so cross-server is later purely additive (more pairs from different `server_id`s + picker UI), no schema or query-shape change.

---

## Testing & gates

All CI gates green locally (mixed frontend + backend PR, rule 09), on top of current `origin/main` (merged; the Discord-banner change was the only conflict — resolved in `authStore.ts` imports):

- `npx tsc --noEmit`
- `npm run lint`
- `npm run dep:check` — no new violations (known-violations baseline unchanged)
- `npx vitest run` — **2586 tests / 364 files** — plus `--coverage` (baseline + frontend hot-path gate, 15/15 hot-path files ≥70%)
- `cargo test --workspace` + `cargo clippy --workspace --all-targets -- -D warnings`

Coverage of the new logic includes: sargable scope narrowing; multi-lib `IN`; FTS-first plan shape (EXISTS, no `JOIN track` before bm25); dedup collapse + priority winner; empty-key rows staying individual; duration guard; backfill correctness + idempotency; migration idempotency across reopen; locale folding (ß/æ/œ/ș/ț/ё/й) with CJK left intact; and the `norm_version`-mismatch rebuild via the read path.

`CHANGELOG.md` (`## Added`) and `src/config/settingsCredits.ts` are updated in this branch.


cucadmuh:
Thanks for the thorough review — addressed everything in `0fe0e074` (branch pushed). All local gates green: `cargo test --workspace` + `cargo clippy --workspace --all-targets -D warnings`, `tsc`, `eslint`, `dependency-cruiser` (no new violations), full `vitest run`, and `test:coverage` (baseline + hot-path).

**Blocking**
- **B1** — `lookup_album_key`/`lookup_artist_key` now read the key as `Option<String>` and flatten, so a NULL identity key opens the detail view un-merged instead of erroring.
- **B2** — `ensure_cluster_keys_built` now runs before the Layer-1 same-server path too (genre browse, advanced search, and the `library_scope_list_albums`/`list_artists` browse commands), so >1-library dedup is correct on a cold index.
- **B3** — `setMusicLibraryFilter` writes the ordered selection as well (`'all'` → `[]`, a folder id → single-entry selection), so it's no longer a no-op once a selection exists.
- **B4** — offline suspend/restore now snapshots and restores the ordered selection (forces `[]` while offline), with a new multi-library regression test.
- **B5** — the `setMusicFolders` prune path bumps `musicLibraryFilterVersion` and defers the catalog reload.
- **B6** — `search.indexIncompleteBanner` and the sidebar picker strings (`librarySelect`/`libraryDeselect`/`librarySelectionCount`) added to all 13 locales.

**Should-fix**
- **S1** — sidecar attach degrades: delete-and-recreate on failure, then continue without it (multi-library dedup off) rather than blocking library open.
- **S2** — read-only cluster `ATTACH` now uses a well-formed `file:` URI (forward slashes, `file:///` for Windows drive paths, query-char percent-encoding) bound as a parameter.
- **S3** — swap / restore / import delete the fixed-name `library-cluster.db` so keys rebuild against the new content instead of stale identity keys.
- **S4** — both lossless browse paths use the sargable `library_id` filter (consistent with the migrated paths) and honor a multi-library selection (new `libraryScopes` field, threaded through the DTO + frontend). Note: lossless still groups by `album_id` (no identity dedup), so the rare same-album-in-two-libraries case can show twice — deferred.

**Minor**
- **M1** documented the 5 s bucket semantics; **M5** removed the 7 dead English-only sidebar strings; **M9** streams the rebuild row-by-row (no whole-table `Vec`); **M10** `ANALYZE track;` after the backfill; **M11** added lossless invoke-payload regression tests.
- **M7** left as-is (generated positional binding always sends both args; harmless `null == absent`). **M2/M3/M4/M6/M8** acknowledged as by-design / follow-ups.

PR #1253, opened 2026-07-07
fix(library): prune orphaned artist/album rows after resync

cucadmuh:
## Summary

Fixes the reported bug where an artist that was played then **renamed** persists forever: it stays in the Artists list (clicking it → "Artist not found"), the new name only appears after an app restart, and further resyncs/restarts don't help.

Root cause — two independent gaps:

1. **Rust:** A full resync soft-deletes orphaned *tracks* (`sweep_resync_orphans`), but the `artist`/`album` browse tables were only ever upserted, never pruned (except a full server purge). Subsonic servers derive artist/album ids from tags, so a rename mints a **new** id and drops the old one → the old browse row lingers. The Artists list (default album-credit mode) reads `artist` rows, but artist detail resolves from live tracks by `artist_id` → stale row → "Artist not found".
2. **Frontend:** The online artist/album browse catalog is cached in module-level maps that were only evicted on library-filter changes, never on `library:sync-idle` → the new artist only appeared after a full module reload (app restart).

## Changes

- **`orphan_cleanup` module** (new): prunes `artist` rows the latest `getArtists` pass no longer confirms (older `synced_at` than the freshest row for that server) that also have **no live track** crediting them, and `album` rows with no live track and no favourite (`starred_at` preserved). Correlated per-server so one statement works scoped (sync) or global (open reconcile). Nothing browsable/openable is dropped.
- Prune runs on **every** artist-index refresh (full sync IS-4 **and** delta DS-9) and again **after the IS-7 track sweep** — so both stable-id and fresh-id-on-rename servers are covered, on full **and** partial (delta) resync.
- **One-time reconcile** at DB open (`library_data_migration`-guarded) clears ghosts already accumulated in existing DBs.
- **Frontend:** online artist/album browse re-keys on the library sync revision, and the buffered catalogs are evicted on `library:sync-idle` — renamed/pruned artists surface without an app restart. Prefetch keys threaded from the app layer to keep `src/lib` off `src/store` (dep-cruiser floor).

Note on the "more cache-flush mechanisms" ask in the issue: rather than add manual per-entity flush buttons, this makes the *automatic* sync path reliably delete orphans and refresh browse — which is the underlying need. Happy to add manual flush controls in a follow-up if still wanted.

## Test plan

- [x] `cargo test --workspace` — new `orphan_cleanup` unit tests (renamed-away artist pruned; live/confirmed/other-scope artists kept; orphan album pruned but starred/live kept; scoped vs global), integration test `full_resync_prunes_artist_orphaned_by_rename`, store reconcile test (prunes once, idempotent/guarded).
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `npx tsc --noEmit`, `npm run dep:check`, eslint on touched files
- [x] `npx vitest run` — new sync-idle cache-eviction test + artist/album feature suites (85 tests)
- [ ] Manual: play an artist, rename it in tags, resync → old artist gone from list, new artist visible without restart; existing DBs cleared on first open after upgrade.

### Library DB migration
- [x] No schema change (additive DELETEs only); no `LIBRARY_DB_SCHEMA_VERSION` bump
- [x] Data cleanup guarded by `library_data_migration` (`orphan_browse_rows_reconcile_v1`), one-time at open
- [x] Ongoing prune wired into sync paths, not run unconditionally on every open

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,7 +180,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Library — renamed artists/albums no longer linger as ghosts after resync
 
-**By [@cucadmuh](https://github.com/cucadmuh), PR [#1253](https://github.com/Psychotoxical/psysonic/pull/1253)**
+**By [@cucadmuh](https://github.com/cucadmuh), reported by Seraphim on the Psysonic Discord, PR [#1253](https://github.com/Psychotoxical/psysonic/pull/1253)**
 
 * Renaming an artist (or album) on the server no longer leaves a stale entry in the local Artists/Albums list that opened to "Artist not found" — a sync now prunes browse-index rows the server no longer lists that also have no remaining tracks, keeping starred albums. Cleanup runs on both full and delta syncs, and a one-time pass at startup clears ghosts already accumulated in existing libraries.
 * The renamed artist/album now appears right after a resync instead of only after an app restart: the Artists and Albums pages refresh their cached catalog when a library sync finishes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The **Settings → Integrations → Discord → Cover art source → Server** option sent your server's cover URL to Discord, which republishes external image links, so anyone viewing your Rich Presence could read your username and login token. That option has been removed — cover art now comes only from **None** (app icon) or **Apple Music**, neither of which carries your credentials. Any saved **Server** setting is switched to **None**. Reported by lavioso on Discord.
 
+### Library — renamed artists/albums no longer linger as ghosts after resync
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#1253](https://github.com/Psychotoxical/psysonic/pull/1253)**
+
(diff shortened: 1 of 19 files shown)


cucadmuh:
Thanks for the review — all addressed in `4de7c833`.

- **B1 (fixed):** `apply_artist_index` now returns the `upsert_index` count, and the prune only runs after a confirmed pass (`count > 0`). An empty/partial `getArtists` no longer drives a prune even though `backfill_from_tracks` advances `MAX(synced_at)`. Added a regression test (`full_resync_empty_get_artists_keeps_album_artist_rows`).
- **B2 (fixed by removal):** dropped the album prune entirely. Agreed the zero-track album-row state is reachable — on-demand inserts (cover backfill, advanced search), partial-but-`Ok` ingest, and the global open reconcile all produce legitimately track-less rows, and there's no `getArtists`-equivalent to supply a freshness stamp. Album-ghost cleanup needs its own positive-confirmation signal; deferring to a follow-up rather than shipping an unguarded delete.
- **S1 (fixed):** added `idx_artist_synced(server_id, synced_at)` via wired migration 018; the scoped prune now computes `MAX(synced_at)` once and does an index seek + range.
- **S2 (fixed):** prune runs once per sync — full sync only post-sweep (IS-7), delta only in DS-9. No more IS-4/IS-7 double run.
- **S3 (noted):** documented the all-libraries scope assumption in the `orphan_cleanup` module doc.
- **N1/N2 (fixed):** `\0syncrev:` format single-sourced in `browseCatalogSyncKey.ts`; clarified the sync-idle clear is memory reclamation complementary to the revision re-key.

Green locally: `cargo test --workspace`, `cargo clippy --workspace --all-targets -D warnings`, `tsc --noEmit`, `dep:check`, eslint, vitest.

PR #1262, opened 2026-07-09
fix(sync): use formPost for large savePlayQueue to avoid HTTP 414

norperz:
## Summary

Large play queues (e.g. playing a big favorites list) fail to sync to the server when paused. Behind Nginx (and similar reverse proxies) `savePlayQueue` is sent as a giant GET with one `id=` per track, which hits the default ~8 KB request-line limit -> **HTTP 414**. The push fails silently, then idle auto-pull re-applies the last successful server queue, so the player looks like it “reverts” to an older track/position. Previously discussed this issue in the [discord](https://discord.com/channels/1490431951348432976/1521575438399439038) with a maintainer.

### Fix

- When the server advertises OpenSubsonic **`formPost`** (Navidrome does), send `savePlayQueue` as `POST` `application/x-www-form-urlencoded` with a short path-only URL (auth + ids in the body).
- Otherwise keep GET; if the proxy returns **414**, retry once as form POST.
- On push failure, suspend idle auto-pull so a failed sync cannot rewind local state; only resume idle pull after a **successful** push.
- `subsonicClient.ts` - shared param serializer + form POST helper
- `subsonicPlayQueue.ts` - formPost / 414 routing
- `queueSync.ts` - failed-push idle suspension + resume-only-on-success

No Tauri boundary change, no settings/on-disk layout change, no locale change. Small queues and servers without a tight URI limit behave as before.

## Test plan

- [x] `npx vitest run` - formPost / 414 / idle-pull cases (new + updated); full `npm test` green (376 files / 2680 tests)
- [x] Manual: large queue (1000+ tracks) on Navidrome behind Nginx -> pause -> access log shows `POST /rest/savePlayQueue.view` (short URI), no 414; queue/position does not snap back after ~30s idle
- [x] Manual: small queue still syncs normally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -209,6 +209,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * The precomputed `library-cluster.db` identity keys used for cross-library dedup are now pruned on rebuild when their track no longer exists (removed, or dropped when a server mints a fresh id on rename). Previously the rebuild only refreshed live tracks and never deleted stale rows, so the sidecar grew with library churn until it was recreated wholesale (server switch / restore / import). The rows were inert (reads only ever join live tracks), so dedup and browse results are unchanged — this just stops the sidecar from bloating.
 
+### Sync — large play queues no longer revert after pausing
+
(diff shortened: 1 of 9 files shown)


cucadmuh (dismissed):
LGTM, further improvements in future PRs.

cucadmuh (approved):

cucadmuh (approved):

PR #1292, opened 2026-07-13
fix(library): sort albums by the artist the row actually shows (#1217)

### LINKED ISSUE #1217: [Bug]: Incorrect "Artist - Year" sorting behavior
### What's wrong?

The "Artist - Year" sorting fails if an album contains tracks with featured artists, even if the album's artist is the same. See the screenshots: screenshot 1 shows the same artist, and he is listed as the only album artist in the tags of all the albums shown. The entire first row in screenshot 1 is arranged correctly: these are completely solo albums without feats(for example, the album "gotika," the tracklisting for which is shown in screenshot 2). However, the second row is where the problems begin: all of these albums feature guest artists on some tracks (see screenshot 3 for the album "Kaidan"). And, as you can see in screenshot 1, they don't sort properly. I also confirm that all tracks in each album are assigned the same year (i.e., where it says "2020," all tracks are from 2020; "2019" is the same, etc.)

### Steps to reproduce

1. Have a few albums of artist WITHOUT feats on them
2. Have a few albums of the same artist WITH feats on them
3. Set sorting "Artist - Year"

### What should happen?

The albums will sort correctly

### What actually happens?

Described in "What's wrong" section and showed on screenshots

### Psysonic version

1.49.0-rc.2

### Operating system

Windows

### Install source

.msi installer (Windows)

Navidrome

### Server version

0.62.0

### Logs

```shell

```

### Screenshots / video

Image

Image

Image

### Anything else

_No response_

Psychotoxical:
@
Closes #1217.

The album browse ordered by `MAX(t.artist)` — the raw **track** artist — while the row mappers derive the displayed artist with `pick_album_group_artist`, which prefers the **album** artist. On an album with featured guests the two differ ("Alpha feat. Zulu" vs "Alpha"), so the album sorted under a name the user never sees and fell out of its artists year run, sometimes landing behind an entirely different artist.

## Summary

- Order by the same rule the row displays, through a shared SQL expression (`sql_display_artist_from`) that each query shape feeds its own columns into: aggregates for the grouped album browse, projected columns for the multi-library dedup path. The `album a` shape already sorts on `a.artist` (the album artist) and is untouched.
- The dedup path built its `ORDER BY` by string-replacing `MAX(t.x)` out of the grouped SQL. That only held while every sort key was a bare aggregate and would have silently mangled the new expression, so it now builds directly from its own columns.
- **Scoped browse (`scope_merge`)**: it feeds one ORDER BY into three query shapes, and the two `GROUP BY t.album_id` ones selected the sort columns unaliased — so `artist` / `album_artist` resolved to bare table columns, taken from an arbitrary row of the group, rather than to the `MAX()` aggregates the row mapper reads. The defect survived on that path. The sort columns are now aliased so the ORDER BY binds to the aggregates.

## Test plan

- Two regression tests, one per path (plain browse and scoped browse). Both use a featured-guest album plus a second artist that sorts between the two spellings. Each was run against the pre-fix expression first and fails there — `[Early, Late, Other, Featured]`, the featured album behind the other artist — and passes with the fix.
- A unit test pins that both order-SQL shapes key on the display artist and that the dedup shape emits no aggregates.
- `cargo test --workspace` — 1053 passing. (`upsert_500_rows_completes_well_under_perf_budget` is a timing-sensitive perf test that can trip under full parallel load; it passes isolated and does so on unmodified `main` too.)
- `cargo clippy --workspace --all-targets` — no findings in the touched crate.
@


Psychotoxical:
Thanks — F1 was right, and the alias commit I pushed before it was not. Fixed in `6e11922f`, along with F2 and F3.

I checked the SQLite semantics directly rather than take the claim on trust, and both halves hold:

| ORDER BY term | Binds to |
|---|---|
| `artist COLLATE NOCASE` (plain identifier) | the **result alias** — `MAX(t.artist) AS artist` |
| `CASE WHEN … artist … END` (identifier inside an expression) | the **table column** `t.artist` |

So an alias is substituted only when the whole ORDER BY term is a plain identifier. Inside the display-artist CASE the bare name resolves against `track`, and a bare column in a grouped query is read from an arbitrary row of the group — I confirmed the arbitrary pick too (it took the group's last row in one shape, and did not follow insert order in another). Aliasing the select list never fixed the CASE form; it only looked fixed.

That also means the defect is nondeterminism, not a stable mis-sort: the sort key depends on which row SQLite happens to pick, so it can change with the query plan, the data, or an SQLite upgrade.

**F1** — split the parameter. `list_albums_layer1_filtered` now takes `grouped_order_sql` and `deduped_order_sql`: the two `GROUP BY t.album_id` branches get the grouped key (aggregates inside the CASE), and only the dedup subquery, which really does project plain columns, gets the deduped one. The other two callers (`list_albums_scope`, genre multi-scope browse) build plain-identifier keys, which alias-resolve correctly in either shape, so they hand the same string to both — no behaviour change there, as you noted.

**F2** — added the multi-track album with a sparse `album_artist`. Worth flagging honestly: **it passes on the unfixed code too**, because the arbitrary row SQLite picked happened to be the favourable one. A behavioural test cannot reliably pin this down — the row containing the MAX value always yields the correct key, so any dataset can get lucky. So I also added a **deterministic guard** asserting the grouped sort key carries the aggregates and leaves no bare column behind. That one cannot pass by luck, and it is what actually holds the line.

**F3** — extended the parity test to `sql_display_artist_from("MAX(t.artist)", "MAX(t.album_artist)")` over **multi-row** groups (a single row cannot tell an aggregate apart from a bare column), covering sparse, absent, blank and compilation album artists.

`cargo test --workspace` green (518 in `psysonic-library`), `cargo clippy --workspace --all-targets` clean. The `upsert_500_rows_completes_well_under_perf_budget` flake you noted showed up once under parallel load and passed on re-run, unchanged by this diff.

The branch is 8 commits behind `main`; I'll rebase before merge.

cucadmuh (approved):
lgtm

PR #1326, opened 2026-07-20
feat: add true simultaneous multi-server support

cucadmuh:
## What this PR delivers

Psysonic can now use several configured Subsonic/Navidrome servers **at the same time as one library**.

This is not only a multi-select filter. The previous product model still allowed a global active server to leak into browsing, mutations and playback. This PR carries server ownership through the full stack: SQLite identity, scoped reads, navigation, UI rows, context actions, queues, streams, covers, lyrics, ratings, playlists, offline/device sync and Orbit.

Users can select music folders from different servers, browse their combined catalogue, enqueue tracks from all of them together and continue interacting with each item on the server that actually owns it.

### One catalogue across several servers

- The sidebar scope is an ordered list of **server + music-folder** pairs, not a folder list tied to one active server.
- **Home, Albums, Artists, Composers, Genres, Favourites, Playlists, Folder Browser, Search, Most Played, Statistics, album details and artist details** aggregate all selected servers.
- Shared albums, artists and tracks are de-duplicated according to scope priority, while every concrete source remains available underneath the logical result.
- A temporarily unavailable or still-indexing server no longer prevents the ready servers in the same scope from loading.

### Mixed-server playback

- One queue may contain tracks from multiple servers; every item resolves its own credentials, stream URL, cover, lyrics, ReplayGain facts and analysis data.
- Play-queue pull/push and restart reconciliation run per owner server, preventing one server snapshot from replacing another server's tracks.
- Gapless, crossfade, shuffle, infinite queue, history, radio and queue restore preserve ownership through queue mutations.
- If one physical copy cannot play, the playback alternative dialog can offer equivalent copies from other selected servers.

### Correct server for every action

- Logical entities retain their concrete server sources, so identical Subsonic IDs from different servers no longer collide.
- Album/artist/track source pickers expose the available physical copies when the user needs an explicit choice.
- Ratings, favourites, playlists, context-menu actions, sharing, offline pins, device sync, covers and metadata mutations target the item's owner instead of the active server.
- Destination-sensitive flows such as playlist creation and smart-playlist editing select the target server explicitly.
- Orbit host/guest state, suggestions, queue operations and cleanup remain attached to the session server even while browsing another server.

### Ownership and scope model

The frontend and Rust library now distinguish:

- a **logical entity identity** used to de-duplicate browse results;
- a **concrete owner/source** used for network requests, writes and playback;
- an **ordered browse scope** of server/library pairs used to rank equivalent sources.

Thin rows and queue snapshots carry stable owner keys through navigation and persistence. Server-specific API calls resolve credentials from that owner rather than reading one mutable global server.

### Indexed multi-server browse

The library database adds migrations `019`–`025` for:

- mainstage feed indexes;
- album, artist, track, starred and composer browse projections;
- folded artist-name search;
- durable identity invalidation.

Combined catalogue reads happen in indexed SQLite queries instead of repeatedly fetching and merging complete server responses in React. Home feeds, text search, details, genre counts, Most Played and Statistics use local scoped reads when ready, with network fallback for servers that are not indexed yet.

### Incremental identity maintenance

Cross-server identity keys and browse projections are maintained incrementally after sync. Exact track/album/artist invalidations are journalled durably, closure queries are indexed, and interrupted work resumes after restart. Startup and foreground refreshes therefore avoid unconditional full-catalogue identity rebuilds.

### Independent server lifecycle

- readiness, reachability, sync progress and sessions are tracked per server;
- full/delta sync preserves ownership through remaps, tombstones and deletions;
- no-op sync avoids unnecessary UI refreshes;
- unavailable servers are excluded from live reads without erasing their configured scope;
- database import/restore reopens through the normal migration and repair pipeline.

## Tauri contract

Boundary changes are additive and generated through tauri-specta. Notable additions/extensions include:

- `library_scope_statistics`;
- `library_scope_most_played`;
- `library_resolve_entity_sources`;
- owner-scoped rating cache DTOs;
- multi-scope genre-count arguments;
- server-bound sync sessions and scoped sync status;
- source metadata on affected browse and playback DTOs.

Rust handlers, generated bindings, TypeScript callers and Vitest Tauri mocks were updated together. No existing command or event was intentionally removed.

## Startup handoff

The startup splash now remains visible until the first route content is committed. Mainstage waits for its initial snapshot/New Releases/Recently Played readiness, while login and lazy routes signal readiness after commit. The splash and app root switch atomically instead of exposing a blank or partially rendered shell.

## Compatibility and migration

- Schema changes are additive; `LIBRARY_DB_SCHEMA_VERSION` and the embedded migration list advance through migration `025`.
- Existing library rows are preserved and upgraded through the standard open pipeline.
- Large reconciliation work is guarded and incremental rather than repeated on every open.
- Server/index-key migrations rewrite persisted frontend ownership keys so existing installs keep their saved state.

## Verification

Local checks on the current branch:

- `npm ci`
- `npm run lint`
- `npm run dep:check`
- `npm test`
- `npx tsc --noEmit`
- `npx vitest run --coverage`
- `bash scripts/check-frontend-hot-path-coverage.sh` — 15/15 files passed
- `git diff --check origin/main...HEAD`

Live migration/restart audit on a 222,376-track catalogue:

- pending identity invalidations: `0`;
- missing cluster keys: `0`;
- orphan cluster keys: `0`;
- identity sidecar schema reopened at version `1`;
- measured 13-row identity closure: about 35 ms, previously about 1.2 s;
- measured 64-row identity closure: about 8 ms, previously about 3.0 s.

GitHub Actions runs frontend tests, frontend coverage/hot-path gates, ESLint, dependency-cruiser, TypeScript, Rust tests, clippy, Rust coverage/hot-path gates and the aggregate `ci-ok` check.

## Review guidance

The diff is large because server ownership is a cross-cutting invariant: fixing only browse UI would leave playback and mutations unsafe, while fixing only API calls would leave persisted queue/navigation identities ambiguous. Commits are split by subsystem and progressively move all affected flows onto the same ownership model rather than introducing parallel implementations.

The application PR is intentionally left open for a human maintainer to merge. Auto-merge is not enabled.


Psychotoxical:
Reviewed at `f8799228` against merge base `c7a01e7d`. Full write-up with all findings, corrections and the verified-clean list: `internal/collaboration/handoffs/2026-07-20-pr-1326-review.md` in workdocs.

**Recommendation: request changes** — four blockers. The architecture holds up and the migration discipline is clean; these are concrete defects, three of them near-one-line.

### 1. Scoped album browse can permanently skip its backfill

`browse_projection.rs:390` (and `is_ready`, `:441-448`) infers "fresh install, nothing to backfill" from `album_browse_projection` being non-empty. Sync also populates that table — `repos/track.rs:125`, `:215`, `:288`.

On an existing install, migration 020 creates the table empty; if a delta sync writes one projection row before the UI calls `library_scope_browse_projection_inspect`, then `migration_started == false && has_projection == true`, so `inspect` returns `needed: false` and `is_ready` returns `true`. All Albums is then served from a projection holding only the albums touched since the upgrade. The completion marker is never written, so there is no self-heal path.

Suggestion: write the completed marker explicitly at DB-creation time when `track` is empty, instead of inferring it from a table sync also writes.

### 2. Upgrade key rewrite deletes offline pins on key collision

`rewriteFrontendStoreKeys.ts:40-43` (and `:61-64`): the `delete albums[key]` sits outside the `if (!albums[nextKey])` guard, so on collision the entry is dropped rather than merged.

Collision is reachable: `serverIndexKeyFromUrl` strips the scheme (`serverIndexKey.ts:8`), so `http://host:4533` and `https://host:4533` — or any two profiles for one host — map to the same key. Result: pinned offline albums vanish and `useLocalPlaybackStore` loses the `localPath` rows, leaving downloaded files on disk with no UI path to reclaim them.

This runs on the upgrade path for every existing install and currently has no test coverage — `rewriteFrontendStoreKeys.test.ts` covers only the `...ForRemap` sibling.

### 3. Playback no longer auto-skips a failing track

`audioEventHandlers.ts:539-557`. Base did:

```js
setTimeout(() => {
  if (getPlayGeneration() !== gen) return;
  usePlayerStore.getState().next(false);
}, 1500);
```

Head ends at `reportPlaybackSourceFailure(...)`, and `playbackAlternativeStore.ts` has no `next()` on any path. When the lookup resolves to `status: 'ready', sources: []` the queue stops on the failed slot.

Since `audio:error` is the engine channel (mid-stream 401/404/reset), unattended playback now halts on one bad track instead of skipping past it. This regresses single-server installs too, not only mixed queues. If halting is intended, the empty-alternatives case still needs a fallback.

### 4. `patchCachedTrack` without an owner patches every server's entry

`queueTrackResolver.ts:294-303` falls back to `key.endsWith(':' + trackId)` when `serverId` is omitted, and `replayGainMetadataSync.ts:87` omits it. The star/rating sites pass it correctly (`pendingStarSync.ts:104`, `:117`).

With small per-server track ids, a queue holding A:`300` and B:`300` gets B's cached title, duration and ReplayGain overwritten by A's metadata upgrade — wrong row text, and wrong gain applied when B plays. `invalidateQueueResolver` (`:279-288`) carries the same unscoped match; currently uncalled but exported.

---

Also worth a look, detail in the handoff: deduped album `song_count` double-counts across servers (`scope_merge.rs:699`, `:710` use `COUNT(*)` where `:459` correctly uses `COUNT(DISTINCT track_dedup)`); the device-sync persist `migrate` drops every legacy source with no notice and the manifest recovery path is closed too; `switchActiveServer.ts:32-41` still tears down a live Orbit session on any active-server switch, with a rationale comment this PR has made obsolete.

Two things that came up during review and did **not** hold, so nobody spends time on them: the startup splash does not hang on an unreachable server (`homeFeedLoader.ts:231-247` races a 4s deadline that resolves with a fallback, and `Home.tsx` clears `loading` in a `finally`), and the `ORDER BY`-alias-over-`GROUP BY` trap does not reproduce against the bundled SQLite — `scope_merge.rs:260-278` is correct, only the explanatory comment at `:426-429` states the wrong reason.

Verified clean, for the record: migration rule passes on all six mandatory points (schema version 25 matches head, every `.sql` wired, strictly additive, journalled and resumable backfill, swap/restore reopen through the migration pipeline, min-compatible unchanged); no files deleted; 90 new test files; i18n complete across all 14 locales; no command or event removed and the event-rate rule is satisfied; offline on-disk layout is collision-free with no mass re-download on upgrade; gapless and per-owner queue push are correct.

cucadmuh:
Addressed the review findings in `92954fbb`:

- Projection readiness no longer treats a partially populated table as complete. Migrations 020/024 now write explicit completion markers only when the live track table is empty, with fresh-database and partial-projection regressions.
- Frontend key migration now matches full profile UUID prefixes (including URL ports) and merges collisions. Offline pins union `trackIds`; local playback keeps the more durable tier and newest playback metadata. Public upgrade-path collision tests were added.
- Playback failure fallback is restored for both `audio:error` and `audio_play` rejection when alternative lookup is empty or errors. Delayed skips are guarded by play generation and the exact owner-qualified queue item.
- Queue cache patch/invalidation is now owner-required. ReplayGain metadata refresh verifies the exact queue ref after the async lookup.
- Deduped album aggregates count logical track winners once across selected servers.
- Legacy device-sync selections are quarantined in persisted v2 state and recovered only after an explicit server-owned selection; ownerless manifests require an explicit fallback owner.
- Active-server switching now preserves the source-bound Orbit session instead of tearing it down; the lower-layer runtime types no longer import the Orbit feature barrel, reducing the dependency baseline from 669 to 581.

Local verification passed: frontend lint, dependency-cruiser, TypeScript, full Vitest and coverage/hot-path gates; Rust workspace tests and clippy; `git diff --check`. GitHub CI is running on the pushed commit.

Psychotoxical:
Re-checked at `92954fbb`. **All four blockers are cleared** — each re-read in the diff.

1. `browse_projection.rs` — "fresh install" now comes from `SELECT NOT EXISTS(SELECT 1 FROM track WHERE deleted = 0)` instead of the projection table that sync also populates, with `partial_incremental_projection_does_not_imply_completion` covering it. Exactly the right shape.
2. `rewriteFrontendStoreKeys.ts` — `mergeOfflineAlbum` / `mergeLocalPlaybackEntry` merge on collision and the delete is conditional on the key actually changing; the function now has direct test coverage.
3. `audioEventHandlers.ts` — auto-skip restored and additionally guarded on `sameQueueItemRef`, so it can't skip past a slot the user or the alternative-source flow already left. Stricter than the pre-PR behaviour.
4. `queueTrackResolver.ts` — making `serverId` required on `patchCachedTrack` and `invalidateQueueResolver` kills the unscoped path structurally rather than at the call site. Best possible fix for that one.

Also picked up: deduped `song_count` (with two-server assertions), the device-sync source handling, the progress-event collision, and the Orbit teardown removed from `switchActiveServer.ts` with a scenario test — so the "session survives moving between servers" claim in the description now actually holds.

Nothing looks papered over: `.dependency-cruiser-known-violations.json` went from 669 to 581 entries, so the baseline shrank rather than absorbing new violations.

### Still open

Findings 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 20, 21 from the handoff are untouched. Four of them write persisted state incorrectly, which is why I'd flag them ahead of the rest — the others are fine as follow-ups:

- `runArtistDetailActions.ts:37` + `ArtistDetail.tsx:108` — a rating failure against the owner server records the capability downgrade on the *active* server, permanently disabling album/artist ratings on a server that supports them.
- `radioStationIdentity.ts:29-38` — a legacy radio favourite is re-bound to the active server's same-numbered station without comparing name or `streamUrl`, and the rewrite is committed to storage.
- `queueSync.ts:197` — a push that never happened returns `true` and clears the idle-pull suspension, so the next idle tick can overwrite the unpushed queue edit.
- `applyServerPlayQueue.ts:151-169` — surplus local slots are dropped on merge with no undo.

Your call whether those land here or as a follow-up; either is defensible, they just shouldn't be lost.

Handoff updated with the full re-check: `internal/collaboration/handoffs/2026-07-20-pr-1326-review.md`.

Minor, while the branch is still open: `92954fbb` is `fix:` without a scope, making four scopeless commits alongside the six `debug(...)` ones — a type that appears nowhere in `main`'s history.

Psychotoxical (dismissed):

PR #1331, opened 2026-07-22
fix(library): repair multi-server scope fallback and add diagnostics

cucadmuh:
## Summary

- repair legacy empty Library scope state when adding another server, so multi-server browse and New Releases do not silently resolve to no selected servers
- apply the active-server fallback consistently to music-folder discovery and local New Releases reads
- add permanent redacted multi-server diagnostics with persisted basic / verbose depth, exposed in Settings and PsyLab only while Debug logging is active
- subscribe AppShell to detailed reachability transitions only while verbose diagnostics are enabled
- keep single-scope Artist browse responsive while a first-run identity rebuild completes in the background
- cover scope repair, fallback behavior, persisted debug depth, redaction, diagnostic gating, reachability subscription gating, and first-run Artist browse with regression tests

## User impact

Existing profiles that retained an empty `libraryBrowseServerIds` array can browse their configured servers again after adding another server. New Releases and music-folder discovery no longer collapse to an empty result in that state. Artist browse no longer waits for a full identity rebuild on the first launch after upgrading an existing large library.

Advanced diagnostics offer basic level 1 and verbose level 3 without exposing credentials or sensitive URL data. The full reachability snapshot is subscribed only while verbose diagnostics are active, so default AppShell renders do not pay for off-by-default logging.

## Verification

- `npm run lint`
- `npm run dep:check`
- `npm test` (491 files, 3403 tests)
- `npx tsc --noEmit`
- `npx vitest run --coverage`
- `bash scripts/check-frontend-hot-path-coverage.sh` (15/15 files pass)

## Scope notes

- Mixed frontend + Rust library change; no Tauri command, event, or payload contract changes.
- Persisted debug depth is backward-compatible and sanitised during rehydrate; the unused level 2 maps to basic level 1.
- All shipped locales include the debug-depth copy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1331](https://github.com/Psychotoxical/psysonic/pull/1331)**
 
 * Existing profiles that retained an empty Library server selection recover the configured active server when another server is added. Music-folder discovery and New Releases no longer collapse to an empty result in that legacy state.
+* Artist browse no longer waits for a full identity rebuild on the first launch after upgrading an existing large library.
 
 
 ## [1.50.0]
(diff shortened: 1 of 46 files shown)


cucadmuh:
Addressed both Medium findings from the workdocs review in `db4a4cba`:

- gated the full reachability snapshot subscription behind active verbose diagnostics, with an early effect guard and regression coverage
- removed inert depth 2; the persisted contract is now basic `1` / verbose `3`, with legacy `2` sanitised to `1`

Also updated the changelog and PR description. Local frontend gates, coverage, Rust workspace tests, and clippy pass. Ready for re-review once CI completes.

PR #1334, opened 2026-07-22
Add streaming quality cap + show the real transcoded format on now-playing

Manwe-777:
## What this adds / solves

Two small, related things around streaming quality:

**1. A manual streaming-quality setting (new feature).**
There was no way to ask the server to transcode the live stream — you could only play the original. This adds **Settings → Audio → Streaming Quality**: pick a bitrate ceiling (Original / 320 / 256 / 192 / 128 / 96 / 64 kbps) and it's sent to the server as `maxBitRate` on `stream.view`. Useful for saving bandwidth on a slow connection when the source is a large lossless file.

The cap applies **only to the live stream**. Hot-cache prefetch, offline downloads, and loudness/waveform analysis keep pulling the original, and locally cached/offline tracks always play at their stored quality.

**2. The now-playing badge showed the wrong format when the server transcodes (bug fix).**
The quality badge was built purely from the track's stored library metadata, so a server-side transcode (e.g. Navidrome forced transcoding, or the new cap above) still showed the *original* file — e.g. `FLAC · 3149 kbps · 24-bit` while actually streaming Opus.

The audio engine already resolved the true codec at decode time but only logged it. This surfaces it to the UI via a new `audio:format` event, and the badges now show the real transmitted format (`OPUS · ≤320 kbps · 48 kHz`) when a transcode is happening, falling back to stored metadata otherwise. When the server transcodes with no cap and the exact bitrate is unknown, the bitrate is dropped rather than shown wrong. Codec name is resolved from the engine's own codec registry so Opus streams are named correctly instead of `?`.

Applied to every surface that shows a live format line: Queue panel, Now Playing hero, mobile player, and Immersive fullscreen. Transcode detection is conservative and container-aware (m4a→aac/alac, ogg→vorbis/opus) to avoid false positives on original files.

## How to verify
- Enable transcoding on the server, set **Streaming Quality** to e.g. 320 kbps, play a lossless track → it streams transcoded and the badge shows the real codec + `≤320 kbps`.
- Play a cached/offline track → still shows the original format (cap does not apply).

## Notes
- **Server compatibility:** requires a Subsonic server with transcoding enabled (Navidrome-first).
- **Tauri boundary:** one additive change — a new `audio:format` event (codec/sample-rate/bit-depth/lossless). No existing commands/events changed; no bindings regeneration.
- **i18n:** English keys added; other locales left for a follow-up translation PR.
- Tests added for the sanitizer, transcode-detection helper, the cap-only-on-live-stream behaviour, and the Rust codec resolver.


cucadmuh:
Thank you for putting this together and for the detailed explanation. There is a lot of thoughtful work here, especially around codec detection and format presentation.

Psysonic is a desktop client rather than a mobile one, so bandwidth limits are generally less critical for us. We have intentionally chosen not to add a manual streaming bitrate cap, rather than simply postponing it.

Several key player features depend on the original track bytes. This includes waveform and loudness analysis, silence-aware AutoDJ, dynamic crossfade timing, natural fade detection, and smooth manual-skip transitions. Waveform data is functional here, not just a visual seekbar.

To keep these features working correctly, Psysonic would still need to download and analyse the original file even while playing a transcoded stream. That creates duplicate network, decoding, and analysis work, which significantly reduces the benefit of requesting a lower-bitrate stream.

After reviewing the complete playback pipeline, the current implementation also has several issues that need to be addressed:

- Capped live, crossfade, and gapless streams can be submitted as canonical analysis for the original track. This can replace waveform, loudness, and enrichment data with results derived from a low-bitrate transcode.
- The first 16 KB are used as the content fingerprint. A transcode produces a different fingerprint, which can create duplicate analysis variants and overwrite the original `track.content_hash` used by readiness and track-ID remapping.
- The playback cache identity ignores `maxBitRate`. A cached 128 kbps stream can be reused after selecting Original, or promoted into the hot cache under the original file suffix.
- `audio:format` has no track, server, or playback-generation identity, so rapid skips and duplicate IDs across servers can attach the format to the wrong track.
- Gapless successor tracks never emit their resolved format.
- The badge uses the current setting rather than the cap used to open the current stream. Same-codec transcoding such as MP3 to MP3 is also not detected.
- Regression tests are needed for analysis isolation, cache identity, hot-cache promotion, gapless transitions, and event identity.


The safest direction may be to separate decoded-format reporting from the streaming-quality setting. The format reporting can be useful on its own, while a bitrate cap would require a quality-aware playback, cache, and analysis architecture that preserves the original-track contracts.

Manwe-777:
Thanks — this is a really useful review; you've clearly traced the whole pipeline and every issue is real. I dug through the analysis and cache paths and want to propose a way to fix all of them while keeping playback to a **single stream**. The short version: the problems are identity/provenance, not audio fidelity, and the pipeline already gives us the seam. I'm happy to iterate on all of it within this PR.

For context on why I'd like to land this rather than drop it: my own setup streams over **Tailscale**, and a full FLAC stream just isn't viable over that link — playback stalls. Today my only options are forcing transcoding server-side in Navidrome (which is what surfaced the badge bug in the first place) or switching to a different client that lets me pick a format/bitrate. A client-side cap would let Psysonic work on constrained/remote links like this, and I doubt I'm the only one running it over a VPN or a slow remote connection.

**The pipeline already streams once.** Analysis is seeded from the *played* bytes, not a second fetch — `enqueue_track_analysis` computes `md5_first_16kb` and seeds from the same buffer (`analysis_runtime.rs:407`, `compute.rs:245`), and `engine.rs` defers `analysis_enqueue_seed_from_url` while the buffer fills "because playback will seed on completion with the same bytes." So capping doesn't force a duplicate download; the only issue is those capped bytes flowing into records that represent the original file.

**Proposed invariant:** capped bytes may drive playback and in-session functional analysis, but must never write `track.content_hash`, drive readiness/remap, or overwrite original-provenance facts. The one primitive needed is an *effective stream quality* (cap + requested format, or "original") threaded from the audio crate into the seed path — which carries no quality today (`submit_analysis_cpu_seed`/`AnalysisCpuSeedJob`, `analysis_runtime.rs:1546/1101`). Then:

- **Canonical analysis & fingerprint** — `patch_content_hash` (`commands.rs:1619`) currently overwrites unconditionally; gate it (and the `track_fact` upsert, `fact.rs:178`) to skip capped-provenance seeds, so `content_hash`/readiness/remap stay original-only.
- **Cache** — ephemeral promote matches live bytes by id only (`playback_identity`, `helpers.rs:274`) and `resolvePlaybackUrl:84` serves it before checking the cap; tag the entry with the streamed quality and serve only on match.
- **Event identity** — add `trackId + serverId + generation` to `audio:format`, emit on the gapless successor, and carry the quality the stream was opened at (an explicit `format` request also makes mp3→mp3 detectable).
- **Tests** — capped-never-canonical, cache isolation, gapless emit, and event identity under rapid skips + cross-server duplicate ids.

Net effect: one stream; playback and functional analysis work off it; the original-track contracts are never touched by transcoded bytes. A track only ever played capped just carries capped-provenance functional facts that upgrade to original the first time it's played uncapped or pinned.

One thing I'd value your input on: for the **waveform/loudness cache** specifically, do you prefer (a) a `provenance` column feeding the `get_latest_*` selection so original always wins (touches the migration set + `OPERATIONAL_TABLES` + schema-version bump), or (b) simply not seeding that cache from capped streams (smaller, but no session waveform for capped-only plays)?

Also: Navidrome takes a `format` param alongside `maxBitRate`, so I'd fold format selection (mp3/opus/aac/original) into the same request — which also makes the badge exact. Happy to rework in whatever order suits.

cucadmuh:
Thanks for the detailed response. I reviewed both options and checked the relevant Navidrome implementation directly. I propose a third approach that preserves compressed-stream analysis without introducing bitrate-specific analysis variants.

## Product model

I am comfortable calculating waveform and loudness from the compressed stream used for playback. For AutoDJ, silence detection, and crossfade, the difference from analysing the original file does not justify downloading the complete original alongside playback.

At the same time, I want to keep one canonical analysis result for each original track revision, rather than separate rows for Original, 64, 128, 320 kbps, or different transcoding formats.

The proposed model separates:

- the identity of the original content; and
- the bytes used as analysis input.

The content identity comes from the original file. Waveform and loudness may be calculated from the compressed playback stream. All bitrate representations of the same original then resolve to the same analysis row.

This is different from both initial options: it retains capped-stream analysis, but does not model each quality as a separate analysis variant.

## Per-address setting

I propose moving streaming quality out of the global Audio settings and associating it with a specific normalized server address.

A server profile may contain both `url` and `alternateUrl`. Each address needs its own streaming-quality preference and capability state.

The intended UX is:

1. The user adds and saves a server normally.
2. Psysonic performs an authenticated identity probe for each configured address.
3. When a particular address reports `type === 'navidrome'`, the bitrate control becomes available for that address in the saved server's settings.
4. Each address defaults to Original, with no Psysonic-requested bitrate cap.
5. The primary and alternate addresses may use different bitrate settings.
6. Playback uses the setting associated with the endpoint selected by the connect layer.

The control should not appear during initial server creation. It becomes available only after the address has been saved and confirmed as Navidrome.

When an address is edited, its Navidrome identity, Range capability, and bitrate preference should be verified again. If it no longer identifies itself as Navidrome, Psysonic should hide and ignore the previous bitrate setting.

Treating the addresses independently is important even when they point to the same Navidrome instance. A LAN endpoint and a public reverse-proxy endpoint may handle Range requests and other transport details differently.

## Original fingerprint probe

Navidrome provides a cheap way to establish original-file identity.

An explicit request to:

```text
/rest/stream.view?id=<track>&format=raw
```

returns before Navidrome applies player-level or server-level transcoding overrides. The raw path opens the original media file directly and serves seekable streams through Go's `http.ServeContent`.

I verified the same relevant behaviour in older releases, including `v0.55.2` and `v0.58.5`, as well as the current source.

The original fingerprint can therefore be obtained with:

```http
GET /rest/stream.view?...&id=<track>&format=raw
Range: bytes=0-16383
```

`stream.view` is preferable to `download.view`, because downloads may be disabled by the server administrator.

The special Navidrome value is `format=raw`, not `format=original`.

For canonical identity, I propose accepting the probe when:

- the selected address has already been identified as Navidrome;
- the response is `206 Partial Content`;
- `Content-Range` starts at byte zero and describes the requested range;
- the body length matches the advertised range;
- the response is not a Subsonic JSON/XML error.

The Range capability result can be cached by normalized address. This keeps capability state aligned with the actual endpoint and its reverse proxy.

## Analysis flow

For capped or server-transcoded playback, the flow becomes:

1. Fetch the original 16 KiB prefix with `format=raw`.
2. Compute the original `md5_16kb` from that prefix.
3. Analyse the compressed bytes already being downloaded for playback.
4. Plan and store the result under the trusted original fingerprint.
5. Reuse an existing complete result when that fingerprint is already known.
6. Replace the active analysis revision only when the original fingerprint changes.

The canonical key remains:

```text
server_id + track_id + trusted original fingerprint
```

The selected address is intentionally not part of this key. Both addresses may point to the same underlying server and media file. The address controls transport behaviour and bitrate selection, while the original fingerprint controls analysis identity.

This requires the trusted fingerprint to travel through the analysis pipeline separately from the bytes being decoded. The current pipeline derives `md5_16kb` from the analysis input, which would otherwise produce a different identity for every transcoded representation.

The corresponding pipeline changes are:

- analysis dispatch accepts a trusted original fingerprint;
- waveform and loudness decode the compressed playback bytes;
- `ContentHashSink` receives only the trusted original fingerprint;
- unverified compressed hashes do not update `track.content_hash`, readiness, or remap state;
- waveform reads use the active trusted content hash rather than an arbitrary latest row;
- older fingerprint rows are cleaned up when a verified replacement becomes active.

This also covers server-forced transcoding when Psysonic did not request `maxBitRate`. The absence of a client cap is not proof of original provenance, so opportunistic playback analysis still needs the raw identity probe whenever the actual stream may have been transcoded.

If the raw-prefix probe fails, playback can continue, but the compressed stream should not produce canonical database or library writes. A runtime-only result may still be used for the current playback generation if the pipeline supports it safely.

## Audio cache identity

Audio cache identity remains separate from analysis identity.

All bitrate representations share one analysis row, but their encoded audio bytes are different. Playback and completed-stream cache identity should therefore include:

```text
server identity
+ normalized selected address
+ track ID
+ requested format
+ requested maxBitRate
```

Authentication salts and tokens can continue to be ignored.

This prevents a completed 128 kbps stream from satisfying a later request for 320 kbps or Original.

For hot-cache promotion, capped bytes can either remain temporary or use a quality-aware tier with truthful format and bitrate metadata. They should not be promoted into the original tier or stored under the original file suffix.

## Active stream information

The requested cap should be latched when the playback generation is created.

Changing the preference while a track is already playing then affects the next stream instead of immediately relabelling the current one.

The `audio:format` payload should carry:

- the server-qualified track identity;
- the playback generation or equivalent stream identity;
- the cap requested for that generation;
- the decoded format.

This allows the frontend to reject stale events during rapid skips or server changes. The same information should carry into gapless successor state so the decoded format remains correct after a chained transition.

The presentation logic should also recognise same-codec transcoding. MP3 to lower-bitrate MP3 and AAC to lower-bitrate AAC are transcodes even when the codec family does not change.

## Remaining PR scope

The other review findings remain relevant alongside this design:

- quality-aware playback and preload cache identity;
- preventing capped bytes from entering the original hot cache;
- decoded-format propagation across gapless transitions;
- playback identity in `audio:format`;
- latching the cap used by the active stream;
- same-codec transcoding;
- all shipped locale keys;
- changelog and contribution attribution;
- an extraction or follow-up plan for the already large decoder module.

The regression coverage should include:

- capped analysis stored under a trusted original fingerprint;
- bitrate changes not creating analysis variants;
- failed raw Range probes not producing canonical writes;
- server-forced transcoding without a client-requested cap;
- independent settings for `url` and `alternateUrl`;
- failover using the selected address's bitrate preference;
- address edits invalidating previous capability state;
- a cached 128 kbps stream not satisfying Original or 320 kbps;
- capped bytes not being promoted as original files;
- MP3-to-MP3 and AAC-to-AAC downsampling;
- stale format events during rapid track changes;
- decoded format after a gapless transition;
- changing the preference while the current stream remains open.

This is the direction I propose for the revision: one original-backed analysis identity, compressed playback bytes as the analysis input, and streaming-quality controls scoped to each confirmed Navidrome address.

Manwe-777:
This design works for me — agreed on all points. The trusted-original-fingerprint model is a clean resolution: one canonical analysis row per original revision, compressed playback bytes as analysis input, and the `format=raw` + Range probe giving identity without a second full download. Thanks for verifying the raw-path behaviour against the Navidrome releases directly.

Agreed as well on:

- **Per-address scoping** — moving the bitrate control onto each saved, Navidrome-confirmed address (with `url` / `alternateUrl` independent, re-verified on edit) rather than a global Audio setting.
- **Cache identity** including requested format + maxBitRate alongside server/address/track, and keeping capped bytes out of the original tier.
- **Latching the cap per playback generation**, with `audio:format` carrying track identity + generation + cap + decoded format, propagated across gapless transitions.
- Probe-failure semantics: playback continues, no canonical writes.

I'll revise the PR toward this design. I've been building regression tests for the earlier findings first — writing each test against the current code so it fails on the defect, then fixing: stale `audio:format` events during rapid skips and cross-server duplicate ids, same-codec (MP3→MP3) downsampling detection, the latched cap, and a cached 128 kbps stream satisfying an Original request are covered so far, with the gapless decoded-format transition in progress. I'll adjust those tests and the pending changes where the new model supersedes them (fingerprint identity, per-address settings, cache key shape) and extend coverage along your list — including the raw-probe failure path, server-forced transcoding without a client cap, and address-edit capability invalidation.

Will also pick up the shipped locale keys, changelog, and a follow-up plan for splitting the decoder module as part of the revision.

Manwe-777:
The revision implementing the agreed design is pushed (9 commits). How each point landed:

**Trusted-original fingerprint.** New `raw_probe.rs`: for any stream whose URL requests a transcode, the first 16 KiB of the original is fetched via `stream.view?format=raw` with `Range: bytes=0-16383`, accepted only on `206` with a zero-based `Content-Range` matching the body length and a body that is not a Subsonic error envelope. The trusted hash travels through the seed path (`enqueue_track_analysis_trusted` → cpu-seed job → `seed_from_bytes_into_cache` → enrichment → `ContentHashSink`), so waveform/loudness/BPM are computed from the played bytes but stored under the original's identity — one analysis row per original revision, and `track.content_hash`/readiness/remap never see a transcode hash. Probe failure ⇒ the seed is skipped entirely (playback continues, no canonical writes).

**Per-address setting.** `streamQualityByAddress` + `streamFormatByAddress`, keyed by normalized address (primary and alternate independently). The controls render in the Servers tab on saved servers only, gated on the identity probe reporting Navidrome; they never appear during creation. Editing an address drops its prefs (unverified until re-probed); removing a server prunes them. Playback resolves the entry for the address the connect layer selected.

**Format selection.** Auto / MP3 / Opus / AAC sent as `format` alongside `maxBitRate`. A format-only request counts as a transcode for all gating (probe, cache, promotion).

**Active stream information.** `audio:format` now carries the server-qualified track identity, the playback generation, the cap latched from the URL the stream was opened with, and the decoded format; the frontend rejects stale/cross-server events. Gapless successors carry their resolved format through `ChainedInfo` and emit at the transition. Same-codec downsampling (mp3→mp3 under a cap) is detected.

**Cache identity.** `same_playback_target` (completed-stream / preload / chain matching) now requires same track + same host + same requested `maxBitRate`/`format` (auth params still ignored), so a completed 128 kbps stream can't satisfy an Original request. Hot-cache entries record the quality they were fetched at and are only reused on match; transcoded bytes are never promoted into the original tier at all.

**Housekeeping.** All 13 non-English locales carry the new keys; changelog entry added under 1.51.0.

**Regression coverage** (each written against the pre-fix code where the defect existed, confirmed failing, then fixed): capped analysis stored under the trusted fingerprint; two bitrate representations sharing one analysis row; raw-probe URL construction + response-validation contract (short files, error envelopes, non-206, wrong ranges); capped streams never promoted; a cached 128 kbps blob not serving Original; quality/host-aware stream-target matching; stale `audio:format` under rapid skips and cross-server duplicate ids; latched cap vs. live setting; gapless format emit; same-codec detection; address-edit invalidation and per-address pref lifecycle.

**Deferred, flagged for follow-up:**
- *Server-forced transcoding without a client cap*: dispatch keys the probe off the URL's own transcode params, so a server-side forced transcode on an uncapped stream still analyzes under the stream hash. Closing it means probing every Navidrome-identified stream (or comparing decoded codec vs. suffix first) — I'd wire that to the same per-address Navidrome capability state; happy to do it in this PR or a follow-up, your call.
- *`get_latest_*` reads pinned to the active trusted hash + stale-row cleanup*: much lower pressure now that trusted keying prevents variant rows from being created, but still worth doing for pre-existing rows.
- *Failover-address preference integration test*: needs connect-cache manipulation in the test harness; the resolution logic itself is covered by unit tests.
- *Decoder module size*: the probe went into its own `raw_probe.rs` rather than growing `decode.rs`; I'd propose extracting the codec-info/event types from `decode.rs` the same way as the follow-up split.

Verified locally end to end: capped playback over a WAN link (Tailscale) with the badge showing the decoded stream, and the analysis DB confirming a single fingerprint per track with `waveform_cache.md5_16kb` equal to the library `content_hash` for tracks played capped.

cucadmuh:
Thanks for the substantial revision. This is much closer to the agreed design, and I appreciate how thoroughly you addressed the first review. I checked the latest head and reran the full frontend and Rust suites locally. Everything passes, including coverage and hot-path gates.

I still found a few issues that need another revision:

- Server-forced transcoding without client `maxBitRate` or `format` still bypasses the trusted-original probe and may write a transcode fingerprint as canonical identity.
- The raw Range probe reads the body before validating the response, so an ignored Range request may buffer the complete file. It also accepts some truncated or inconsistent `Content-Range` responses.
- The frontend receives `audio:format.generation` but does not validate it or clear the previous resolved format when a new playback generation starts. Same-ID replays and cross-server switches can therefore show stale format data.
- `streamCapKbps: null` is treated as a missing legacy field and replaced with the current setting, which can relabel an uncapped stream.
- Gapless transitions update the URL and format but retain the previous track/server analysis identity.
- A successful trusted probe may not repair the active library hash when analysis is already complete, while reads can still select a newer stale variant.
- The raw probe does not apply configured reverse-proxy headers, and CPU seed jobs still deduplicate by bare track ID across servers.
- Editing a server address retains stale Navidrome capability until reprobe.
- The two new selectors are not fully keyboard operable, and “Original (no transcode)” conflicts with choosing an explicit MP3/Opus/AAC format.

Thank you again for working through a complex part of the playback pipeline. The remaining items are narrower than in the first revision, but the provenance, probe, and playback-identity issues still need to be resolved before approval.

Manwe-777:
All nine findings are addressed in the latest push. Point by point:

- **Server-forced transcoding without client params** — the raw-prefix probe now runs for every HTTP stream, not only URL-marked transcodes. Success stores under the trusted original; failure on a transcode-requesting URL skips the seed; failure on a plain URL falls back to the legacy bytes-hash path, so non-Navidrome servers (where the raw contract doesn't exist and validation rejects) behave exactly as before. No explicit Navidrome gate needed in the audio crate — the strict probe contract is self-gating.
- **Probe body handling** — status and `Content-Range` are validated before any body bytes are read (a 200 that ignored Range is rejected without buffering), the body is streamed under a hard cap, and truncated/inconsistent ranges (`end >= total`, wrong window, `*` total) are rejected. The window check is exact: `end` must equal `min(16 KiB, total) − 1`, since any other prefix fingerprints differently than the original.
- **Generation validation / stale format** — starting a new playback clears the previous resolved format, and events from an older generation of the same track are rejected, covering same-id replays and cross-server switches.
- **`streamCapKbps: null`** — an explicit null is now a real "no cap"; only a truly absent field takes the legacy fallback.
- **Gapless analysis identity** — the transition re-pins the engine's analysis track/server identity to the successor alongside the format emit.
- **Repair + stale variants** — a trusted seed records the hash even when analysis is already complete (repairing a stale `track.content_hash`), and once a verified row is active all rows under other fingerprints are purged, so latest-row reads can't select a stale variant.
- **Probe headers / seed dedup** — probe requests apply the configured reverse-proxy headers via the same mechanism as playback; CPU seed jobs deduplicate and merge by `(server, track)`. The old test that encoded cross-server merging ("latest server scope wins") is replaced by one asserting two servers with the same id stay two jobs.
- **Address-edit capability** — editing `url`/`alternateUrl` now also drops the cached server identity, so Navidrome-gated controls hide until the reconnect probe confirms the endpoint.
- **UI** — `CustomSelect` gained full keyboard operation (arrow/Home/End navigation with a visible highlight, Enter/Space selection, `aria-activedescendant`, scroll-into-view), which benefits every dropdown using it; and the bitrate option was relabelled to "No bitrate cap" in all locales so it no longer contradicts an explicit format choice.

Regression tests accompany each behavioural fix (out-of-order generation, explicit-null cap, cross-server seed scoping, variant purge, probe header/range contract, address-edit identity drop).

One residual I want to flag rather than hide: the HTTP **backfill** worker in the analysis crate downloads via plain `stream.view` and would also see server-forced transcoded bytes; the probe lives in the audio crate and the dependency direction (audio → analysis) means it can't be called from there today. The clean fix is moving the probe module into `psysonic-analysis` so the backfill fetch path uses the same trusted-identity flow — happy to do that here or as a follow-up, whichever you prefer.

Manwe-777:
Heads-up on scope: the read-only decoded-format reporting has been extracted into #1338 as a standalone PR, per the earlier suggestion to separate it from the streaming-quality setting — it's lower-complexity and can ship first. #1338's description carries the full merge-order/rebase plan; short version: once it merges, this branch gets rebased onto `main` (shared hunks are byte-identical and dissolve), leaving this PR as purely the per-address quality controls, trusted-original probe, and cache-identity work. I'll announce here before the rebase force-push in case anyone has local work on this branch by then.

Manwe-777:
Rebasing this branch onto `main` now that #1338 is merged (as planned) — force-push with lease incoming in a moment. Content is unchanged apart from the rebase: the tree was verified to differ from the previous tip only by main's own changes.

Manwe-777:
The latest review round is addressed in four commits (one per area, with the analysis-crate provenance/queue work kept together since those changes interleave in the same functions). Finding → fix → regression coverage:

| Finding | Fix | Regression test |
|---|---|---|
| Backfill producers bypass the probe | Probe moved into `psysonic-analysis`; playback dispatch and both HTTP backfill producers now share one provenance path (`resolve_trusted_identity`) | Request-level tests against a mock server: valid `206` → trusted fingerprint; `200` that ignored Range → rejected before any body read; Subsonic error envelope served as `206` → rejected |
| Plain-URL probe failure still writes canonical identity | Capability-aware policy: a server that has ever answered a raw probe correctly this session is known raw-capable — later probe failures there mean *unverifiable* and skip canonical writes. Servers with no raw contract keep the legacy bytes-hash path, so non-Navidrome behaviour is unchanged | Known-capable server + failing probe → `SkipCanonicalWrites`; unknown server + failing probe → `AssumeOriginal` |
| Trusted purge half-wired; reads take the newest row | The "already complete under the trusted hash" branch now repairs the library hash **and** purges other-fingerprint rows (extracted as `repair_complete_trusted_identity`) | Your exact chain: complete trusted row + newer transcode variant → after repair, `get_latest_md5_16kb_for_track` returns the trusted fingerprint (the test asserts the stale row won the read beforehand) |
| Backfill queue dedups by bare track id; `analysis_track_in_cpu_pipeline` never matches | Backfill queue reserves/merges/finishes by `(server, track)`; the pipeline gate takes the server scope and matches the composite-keyed maps again | Same id on two servers → two independent jobs (was `DuplicateSkipped`); the old cross-server-overwrite test now asserts same-server merge only |
| Cross-server gapless resolves the predecessor's server | `current_playback_url` is updated (and the analysis identity re-pinned) **before** the transition analysis spawns, so URL-derived server resolution sees the successor | No automated test — asserting the resolved scope needs a live engine + AppHandle harness, which was heading toward flaky; happy to add one if you have a preferred harness pattern |
| Byte-cache identity keys on transport authority only | `same_playback_target` compares the full server base (scheme + authority + path prefix before `/rest/`) plus track + quality | Proxy-prefix collision (`/nav-a` vs `/nav-b` on one host) rejected; same prefix with rotating auth still matches; http vs https distinct. Note: this achieves the isolation via URL server-base rather than threading `serverIndexKey` into the cache entries — same guarantees with much less churn through the entry structs; can switch to explicit keys if you prefer |
| Frontend lint gate red on `CustomSelect` | Render-pure `useId`-based listbox ids | Focused keyboard/ARIA suite: arrow/Home/End navigation with disabled-option skipping, Enter selection, `aria-activedescendant` tracking |
| Probe frequency on cached replays / complete tracks | Left as-is per your note; the planner-gating idea is noted for a follow-up | — |

All suites pass locally (frontend gates incl. lint/dep-cruiser; Rust workspace with the usual pre-existing `psysonic-library` timing flakes only). Ready for the minor-issues list whenever you send it.

Manwe-777:
All six findings addressed (three commits — the provenance/queue/scope changes interleave in the same functions, so they ship together; byte-cache identity and the frontend guard are separate). Finding → fix → regression:

| Finding | Fix | Regression test |
|---|---|---|
| First probe failure returns AssumeOriginal | Canonical writes from HTTP-stream bytes now require **positive** original provenance: any probe failure — including a first-ever one — skips canonical writes at both producers. The session capability cache is gone (insufficient: forced transcoding is invisible on the wire) | Fresh server + failed probe ⇒ `SkipCanonicalWrites` (request-level) |
| Spill / promotion lose provenance | `spawn_track_analysis_file` carries the source stream URL, so spilled ranged streams and preloaded HTTP files keep stream provenance; stream-to-local promotion is gated on the promoted bytes (memory or spill prefix) fingerprint-matching a verified raw probe of the original | Prefix-equality gate: original bytes accepted, transcoded bytes rejected |
| Running CPU job swallows a different trusted fingerprint | Seed dedup identity is (server, track, **content revision**) — revision = trusted fingerprint, else the bytes' own; the pipeline gate prefix-matches any revision | Running revision A + submission of revision B ⇒ B queued as its own job with its bytes intact (was `RunningFollower`) |
| Analysis scope prefers the playback URL | Precedence flipped: explicit canonical server key → pinned scope → URL-derived last (pure `resolve_analysis_scope`) | Primary vs alternate address resolve to the one canonical scope |
| Generation guard cleared with the format object; gapless retains predecessor format | Monotonic `streamFormatGenerationFloor` on the player store survives clears; gapless advancement (all callers of `applyGaplessQueueAdvance`, incl. progress reconciliation) nulls the predecessor's format | Post-clear replay rejects the older generation; gapless advance clears a populated format |
| Profiles share encoded bytes on one endpoint | The byte-cache identity signature includes the owning account (`u=`); rotating auth params stay excluded | alice/bob on one endpoint distinct; same account with rotating salts still matches |

One functional consequence to confirm you're comfortable with: strict positive provenance means servers without the raw contract (non-Navidrome) no longer get analysis seeded from HTTP streams at all — their coverage now comes from local files and downloads only. That is the literal reading of the requirement and I think the right call, but it is a behavior change for those servers worth an explicit nod.

All suites pass locally (frontend gates; Rust workspace with only the pre-existing `psysonic-library` timing flakes; clippy clean on touched files).

PR #1335, opened 2026-07-22
fix(library): resolve Various Artists albums to the VA entity

Psychotoxical:
## Summary

- The "Various Artists" entity links its compilations through the `album_artist` label, not a per-track `artist_id`. The local artist page (a scoped, index-first query since the local-first switch) followed track `artist_id` and so found only the handful of tracks literally tagged with the VA id instead of every compilation.
- **Artist detail:** when the target is the VA entity (resolved from the artist row's canonical name, so it works even when no track carries the VA id), union its `album_artist`-linked compilations into the album grid. Non-VA artist pages are unaffected.
- **Album detail:** derive the hero artist id from the album-artist (`raw_json.albumArtistId`) so a compilation links to "Various Artists" instead of a representative track's performer, falling back to the track id when the album has no named album-artist. The JSON1 read is guarded against malformed `raw_json` so one bad row cannot abort the query.
- The album-artist identity is resolved into the hot columns for every album and kept authoritative through the album-row overlay and the frontend `raw_json` merge, so the resolved link is not overwritten by the server's legacy performer id. Backend and frontend use the same word-boundary VA-label predicate so they never disagree on the same unlink.

## Test plan

- `cargo test --workspace` — new regression tests for both bugs, the pure-label and no-double-count union paths, the malformed-json guard, the overlay VA-unlink and feat-name cases, the non-VA-not-seeded and album-count behaviour, and the word-boundary matcher; each proven to fail against the unfixed code.
- Frontend: `tsc`, `eslint`, `vitest` (mapper VA-unlink parity + loader payload) all clean.
- `cargo clippy --workspace` clean.
- Verified against a real multi-server index (label linkage, over-inclusion, and query shape).

## Deferred (own scope)

- Album-artist links on the general browse grids and the non-VA compilation-ish fallback are left as-is here — the same class on other surfaces, out of this fix's scope.
- VA-label detection is a word-boundary match kept identical on both sides, not an exact-string match; tightening the VA semantics is a separate cluster-policy decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Playing a multi-disc album from the header no longer interleaves the discs (disc 1 track 1, disc 2 track 1, disc 1 track 2, …). Tracks queue in disc-then-track order, so disc 1 plays in full before disc 2. Tracks without a disc number are treated as disc 1, matching the track list.
 
+### Artist details — Various Artists shows every compilation
+
+**By [@Psychotoxical](https://github.com/Psychotoxical), PR [#1335](https://github.com/Psychotoxical/psysonic/pull/1335)**
+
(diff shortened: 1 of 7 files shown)


cucadmuh (dismissed):
lgtm

PR #1336, opened 2026-07-23
feat(album): show disc cover in the multi-disc separator

### LINKED ISSUE #1333: [Feature]: Replace CD icon in multi-disc albums with album cover or disc cover
### What's the use case?

Use case for when you're having a look at multi-disc albums(such as Ozzy Osbourne's Live & Loud), Psysonic displays them with separators with a disc icon and CD disc number beside it.

### Proposed solution

I'd like to suggest an improvement for Psysonic's multi-disc album display, based on what Navidrome already does. Replace the CD icon that is displayed beside CD 1/2/3 with the album or Disc image(if there is one) and make it a little larger(like 48x48). Navidrome already displays the album covers.

### Alternatives considered

_No response_

### Anything else

Here are some references:
_How it displays(Psysonic)_

Image

Image

_How Navidrome does_:

Image

_It also displays disc specific artwork if there are properly named pictures._

I don't know how feasible this is, but it would be something cool to have.

PS.: In order to not cause trouble for the user, revert to the current CD icon if Navidrome can't find a disc artwork. Or leave it to the user to choose what to do in case of multi-disc albums, such as use the album artwork for both discs or use the disc icon.

Thanks.

Psychotoxical:
Closes #1333

## Summary

- The multi-disc separator ("CD N") shows each disc's cover instead of a generic CD glyph (desktop + mobile).
- On **Navidrome**, per-disc art uses the canonical `dc-<albumId>:<disc>` cache slot, gated by a real multi-disc count (`library_album_disc_count`). Single-disc albums and non-Navidrome servers stay on the shared album slot.
- The same disc-aware path covers the **queue**, **playbar mini-cover**, album **disc separator**, and **"Who is listening?"** presence rows — so a box set no longer shows only disc 1 outside the album page.
- Presence resolves via `usePresenceCoverRef` (album id + disc) instead of feeding a track's `mf-*` id into the shared album cache slot.
- All Albums cover warm now passes the album id and refuses fetch-only `mf-*`/`tr-*` entity keys, so clearing the cover cache no longer leaves junk `album/mf-…` directories.
- Browse rails stay album-scoped / IPC-free (`libraryResolve: false`), as on `main`.

## Review note — cover pipeline

- Separator uses `DiscHeaderCover` → `useDiscCoverRef` (Navidrome `dc-*`, otherwise track-cover with `forceDistinctDiscCovers` only when the disc track has a usable cover id).
- Track/playback covers use `useMultiDiscScopedCoverRef` only when Navidrome ∧ multi-disc ∧ library resolve — no per-song cache explosion.
- Depth-3 `emitCoverDebug` reports residual mf-album-slot resolve/ensure and warm skips when debug logging is on.

## Test plan

- [ ] Real multi-disc release with distinct per-disc covers: each "CD N" header shows its own cover; queue / playbar / Who is listening? match the playing disc.
- [ ] Single-cover / single-disc albums still reuse the shared album art everywhere.
- [ ] Non-Navidrome server: no `dc-*` requests; separator falls back sanely.
- [ ] Clear cover cache, open All Albums, confirm no new `cover-cache/.../album/mf-*` dirs.
- [ ] `tsc`, cover/presence/warm vitest suites, `dep:check` clean.

## Deferred

- Cold `dc-*` offline no longer falls back to a warm shared `al-*` hero slot (removed mid-PR hand-rolled fallback). A shared, fetch-outcome-aware answer belongs in the cover pipeline if that blank matters in practice.
- Disc-cover box size (40 px) is still duplicated between the component hint and CSS; not tokenized here.


Psychotoxical (commented):
Reviewed the standard-track-cover refactor (bd9e718b). Routing the disc separator through `useTrackCoverRef` unifies its cache slot and fetch id with the track rows on the same page, and the album-level distinct-disc detection (seeded from the full tracklist in AlbumDetail) is the principled source of truth — a genuine box set resolves per-disc while a single-cover album reuses the shared album slot the hero already warmed, so the removed offline fallback has nothing left to cover. No dangling references to the removed helpers; `albumCoverRefForSong` is correctly retained. Verified locally: tsc, eslint and dep:check clean; cover and album vitest suites green. LGTM.

cucadmuh:
Re-reviewed at `9230f0c5` (post-rebase hygiene).

- Rebased onto current `main`; `CHANGELOG.md` conflict resolved (keeps #1335 / #1338 / #1340 from main).
- CHANGELOG bullets expanded for queue/playbar/presence per-disc art and the All Albums `mf-*` warm-dir fix under the same #1336.
- PR description refreshed to match the branch (`useDiscCoverRef` + `dc-*` unify + warm fix; removed stale `useDiscSeparatorCoverRef` / offline-fallback wording).
- `useDiscCoverRef` now uses the index-key-aware `useServerIsNavidrome` helper; credits line added.

Cover design still looks good. Confirm CI green on this head; maintainer merges — I won’t merge.

PR #1337, opened 2026-07-23
test(library): make the cross-server order test prove its tie-break

Psychotoxical:
Follow-up to the post-merge review of #1332.

- The cross-server regression test passed without exercising the final `server_id` key: its tied rows already emerged from the ranked partitions in the asserted order, so dropping the key left the test green.
- The fixture now opposes the assertion on every incidental ordering — the `s1` row sorts after the `s2` one by title/dedup key, and the rows are seeded s2-first.
- Comments corrected: the final ordering is lexical `server_id`, not scope priority. Only the dedup inside `ranked` is priority-driven.
- Test-only; the production query is unchanged.

## Test plan

- `cargo test --workspace album_detail_disc_order_tie_break_is_total_across_servers` — pass.
- Mutation check: removing `server_id ASC` from the `album_detail` `ORDER BY` makes it fail with `[("s1","s1-anchor"), ("s2","dup"), ("s1","dup")]`, i.e. the reverse order. Query restored afterwards.
- `album_detail_orders_tracks_disc_then_track` and `find_by_album_orders_by_disc_then_track_then_id` — pass.
- `cargo clippy --workspace --all-targets` — no findings in `psysonic-library`.

### DIFF
--- a/src-tauri/crates/psysonic-library/src/scope_merge.rs
+++ b/src-tauri/crates/psysonic-library/src/scope_merge.rs
@@ -3706,8 +3706,15 @@ mod tests {
     fn album_detail_disc_order_tie_break_is_total_across_servers() {
         // The scoped loader merges a cross-server album, so a server-local `id` is
         // not a total tie-break: two surviving tracks from different servers can
-        // share disc, track number, and id. `server_id` is the final key, so the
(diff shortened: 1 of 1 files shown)


cucadmuh (approved):
lgtm

PR #1338, opened 2026-07-23
Show the real decoded stream format in now-playing, not stored metadata

Manwe-777:
## What this solves

The now-playing quality badge is built from the track's **stored library metadata**, so when the server transcodes the stream (e.g. Navidrome forced transcoding), the player still shows the original file — `FLAC · 3149 kbps · 24-bit` while actually receiving Opus.

The audio engine already resolves the true codec at decode time but only wrote it to the debug log. This PR surfaces it to the UI via a new additive `audio:format` event: badges on the queue panel, now-playing hero, mobile player, and immersive fullscreen show the **actually decoded** format when a transcode is happening (with the original file's format in a tooltip), and fall back to stored metadata otherwise. When the transmitted bitrate is unknown, none is shown rather than a wrong one.

## Scope

This is the read-only decoded-format reporting split out of #1334, per the review discussion there — it is useful on its own and ships independently of any streaming-quality controls. The event contract already carries the identity fields agreed in that review (track id, playback server, generation, latched `maxBitRate` from the stream URL — always absent today since nothing sends the param), so #1334 rebases onto this cleanly.

- Codec names resolve via the engine's own codec registry (includes the libopus adapter) — Opus streams read `OPUS`, not `?`.
- The frontend drops events for a since-skipped track, a duplicate Subsonic id on another server, or an older playback generation; a new playback clears the previous format.
- Gapless successors emit their resolved format at the transition (a chained advance never re-runs the play command).
- Transcode detection is container-aware (`m4a`→aac/alac, `ogg`→vorbis/opus) to avoid false positives on originals.

## Verification

- Regression tests: stale-event rejection (rapid skips, cross-server duplicate ids, out-of-order generations), explicit-null cap handling, gapless format emit (mock emitter), codec-registry resolution, format/transcode-detection helper cases.
- Full frontend and Rust suites pass locally; verified live against Navidrome with forced transcoding (badge reads the transcoded stream; original shown in tooltip).

Screenshot 2026-07-23 at 4 20 29 PM

## Notes

- **Tauri boundary:** one additive event (`audio:format`). No commands changed or renamed; no bindings regeneration (events are hand-typed).
- **i18n:** the badge tooltip key is included for all shipped locales.
- **Server compatibility:** display-only; works with whatever the server transmits, no server support required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Merge order & rebase plan (this PR vs #1334)

This PR is a strict subset of #1334, extracted so the display-only fix can ship first. Intended sequence:

1. **Merge this PR (#1338) first.** It is self-contained: display-only, no persisted settings, no migrations, no on-disk changes, one additive Tauri event. Safe for any patch release.
2. **Then #1334 gets rebased onto `main`.** The shared code (event contract, badge surfaces, gapless emit, format helper, locale keys) is byte-identical here, so those hunks dissolve in the rebase and #1334 reduces to the streaming-quality feature itself: per-address bitrate/format controls, the trusted-original fingerprint probe, quality-aware cache identity, and the second-review fixes that only make sense alongside the cap feature. I'll do that rebase promptly after this merges and push with `--force-with-lease` — if anyone has local work on the #1334 branch at that moment, flag it there first so nothing gets stranded.
3. **Reviewing order note:** any review feedback on the shared code is best raised here (it lands in both), while cap/probe/settings feedback belongs on #1334.

Release/changelog: the 1.51.0 changelog entry currently sits in #1334 and describes the combined feature set. If this PR ships in a release before #1334, say the word and I'll add a standalone changelog line here (decoded-format display only) and trim the #1334 entry accordingly.

If the order ends up inverted for any reason (#1334 first), this PR becomes redundant and should simply be closed — everything in it is contained there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Playing a multi-disc album from the header no longer interleaves the discs (disc 1 track 1, disc 2 track 1, disc 1 track 2, …). Tracks queue in disc-then-track order, so disc 1 plays in full before disc 2. Tracks without a disc number are treated as disc 1, matching the track list.
 
+### Now playing — show the real streamed format when the server transcodes
+
+**By [@Manwe-777](https://github.com/Manwe-777), PR [#1338](https://github.com/Psychotoxical/psysonic/pull/1338)**
+
+* The quality badge (queue, now-playing hero, mobile player, immersive fullscreen) was built from the track's stored library metadata, so a server-side transcode still read the original file — e.g. "FLAC · 3149 kbps · 24-bit" while actually receiving Opus. It now shows the format the audio engine actually decoded, with the original file's format in a tooltip, and falls back to the stored metadata when no transcode is happening. When the transmitted bitrate is unknown, none is shown instead of a wrong one.
+
 
 ## [1.50.0]
 
--- a/src-tauri/crates/psysonic-audio/src/commands.rs
+++ b/src-tauri/crates/psysonic-audio/src/commands.rs
@@ -347,6 +347,7 @@ pub async fn audio_play(
     let duration_secs = built.duration_secs;
     let output_rate = built.output_rate;
     let output_channels = built.output_channels;
+    let resolved_format = built.resolved_format;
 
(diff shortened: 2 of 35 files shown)


cucadmuh (approved):
lgtm

Issue #1072, opened 2026-06-12
[Bug]: Client becomes unresponsive partway through playback of a song


GrohJann:
### What's wrong?

Whenever I play a song, the app freezes during playback and becomes unresponsive. The song continues to play.

### Steps to reproduce

1. start Psysonic
2. play song
3. wait

### What should happen?

The UI should continue to work, showing the songs progress.

### What actually happens?

UI freezes completely. Song progress and lyrics stop updating.

### Psysonic version

1.47.0

### Operating system

Linux

### Install source

AUR — psysonic-bin

### Subsonic-compatible server

Navidrome

### Logs

```shell
psysonic
Note: disabling nvidia explicit sync.
See https://bugs.webkit.org/show_bug.cgi?id=280210 for more details
[2026-06-12 09:17:35.473 +0200] [psysonic] logind PrepareForSleep listener registered (post-sleep audio recovery)
[2026-06-12 09:17:36.024 +0200] [analysis][waveform] db hit track_id=ABF6BDTrGvnDVCSRMUL4UX bins_len=1000 bin_count=500 updated_at=1781187065
[2026-06-12 09:17:36.160 +0200] [frontend][normalization] {"step":"init:set-normalization","details":{"engine":"loudness","targetLufs":-14,"currentTrackId":"ABF6BDTrGvnDVCSRMUL4UX"}}
[2026-06-12 09:17:36.160 +0200] [normalization] audio_set_normalization requested_engine=loudness resolved_engine=loudness target_lufs=-14.00 pre_analysis_db=-4.50
[2026-06-12 09:17:36.161 +0200] [analysis][waveform] db hit track_id=ABF6BDTrGvnDVCSRMUL4UX bins_len=1000 bin_count=500 updated_at=1781187065
[2026-06-12 09:17:36.162 +0200] [frontend][normalization] {"step":"refresh:start","details":{"trackId":"ABF6BDTrGvnDVCSRMUL4UX"}}
[2026-06-12 09:17:36.240 +0200] [frontend][hot-cache] {"event":"prefetch-auth-off","clearedPending":0}
[2026-06-12 09:17:36.259 +0200] [frontend][normalization] {"step":"refresh:hit","details":{"trackId":"ABF6BDTrGvnDVCSRMUL4UX","row":{"integratedLufs":-16.21401032920018,"truePeak":0.9767438173294067,"recommendedGainDb":-0.7956134174763391,"targetLufs":-14,"updatedAt":1781187065}}}
[...]

PR #1348, opened 2026-07-26
fix(appimage): prefer host WebKitGTK over the bundled copy when available

### LINKED ISSUE #1347: [Bug]: Blank window on Fedora 44 (AppImage) — bundled WebKitGTK aborts with EGL_BAD_PARAMETER
### What's wrong?

The AppImage shows an empty/blank window on Fedora 44 (Universal Blue / Bluefin,
GNOME Wayland, AMD Radeon 680M). The AppImage bundles its own WebKitGTK
(webkit2gtk-4.1, Ubuntu build), which is incompatible with the host graphics
stack (Mesa / libglvnd 1.7): EGL display creation fails and WebKitWebProcess
crashes with SIGABRT on every launch, so the UI never renders.

Extracting the AppImage and running usr/bin/psysonic directly against the
system webkit2gtk4.1 (2.52.5) works perfectly — so the app itself is fine,
only the bundled WebKit is the problem. Tauri docs explicitly recommend NOT
bundling WebKitGTK into AppImages; excluding libwebkit2gtk / libjavascriptcoregtk
/ libsoup (and GL-related libs) from the bundle should fix this for all
Fedora/atomic-distro users.

### Steps to reproduce

1. Download Psysonic_1.50.0_amd64.AppImage from GitHub Releases on Fedora 44 (Wayland session, AMD GPU).
2. chmod +x and run it from a terminal.
3. A window opens but stays blank; stderr shows the EGL error below.

### What should happen?

The login screen renders inside the window.

### What actually happens?

The window stays blank. stderr:

  Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

coredumpctl shows the bundled WebKitWebProcess crashing with SIGABRT on every
launch (executable: /tmp/.mount_Psyson*/usr/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess).

Workarounds that do NOT help: WEBKIT_DISABLE_DMABUF_RENDERER=1,
WEBKIT_DISABLE_COMPOSITING_MODE=1, LIBGL_ALWAYS_SOFTWARE=1, GDK_BACKEND=x11.
What DOES work: --appimage-extract and running usr/bin/psysonic against the
system webkit2gtk4.1-2.52.5 — no crashes, UI renders normally.

### Psysonic version

1.50.0

### Operating system

Linux

### Install source

None

Navidrome

### Server version

_No response_

### Logs

```shell
Launch output (stderr):

    (psysonic:17662): libayatana-appindicator-WARNING **: libayatana-appindicator is deprecated.
    [psysonic] logind PrepareForSleep listener registered (post-sleep audio recovery)
    Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

WebKitWebProcess crash (systemd-coredump, repeats on every launch):

    PID: 17706 (WebKitWebProces)
    Signal: 6 (ABRT)
    Command Line: ././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess 4 24
    Executable: /tmp/.mount_PsysonapmDbl/usr/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess

    Stack trace of thread 17706:
    #0  0x00007fd34e5c8ccc __pthread_kill_implementation (libc.so.6 + 0x74ccc)
    #1  0x00007fd34e56de8e raise (libc.so.6 + 0x19e8e)
    #2  0x00007fd34e5557b3 abort (libc.so.6 + 0x17b3)
    #3  0x00007fd35771361c n/a (n/a + 0x0)
    #4  0x00007fd35576a5df n/a (n/a + 0x0)
    #5  0x00007fd355754a26 n/a (n/a + 0x0)
    #6  0x00007fd35570c928 n/a (n/a + 0x0)
    #7  0x00007fd35570ac37 n/a (n/a + 0x0)
    #8  0x00007fd3554026a6 n/a (n/a + 0x0)
    #9  0x00007fd354c84e0d n/a (n/a + 0x0)
    #10 0x00007fd35511c592 n/a (n/a + 0x0)
    #11 0x00007fd35511c848 n/a (n/a + 0x0)
    #12 0x00007fd35511c998 n/a (n/a + 0x0)
    #13 0x00007fd352882405 n/a (n/a + 0x0)
    #14 0x00007fd352948dad n/a (n/a + 0x0)
    ...
    #21 0x00007fd34e557681 __libc_start_call_main (libc.so.6 + 0x3681)
    #22 0x00007fd34e557798 __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x3798)

    Note: the "n/a" frames belong to the WebKit libraries bundled inside the
    AppImage (no debug symbols available on the host). The modules systemd DID
    resolve are all host Fedora RPMs (libEGL/libGLX/libGL from libglvnd-1.7.0-9.fc44,
    libgbm from mesa-26.1.4-4.fc44) — i.e. the bundled WebKit is mixing with the
    host GL stack, which is exactly where it aborts.

    No such crashes occur when the extracted usr/bin/psysonic runs against the
    system webkit2gtk4.1-2.52.5-1.fc44.
```

### Screenshots / video

_No response_

### Anything else

Host: Bluefin (Fedora Silverblue 44 based), GNOME on Wayland, kernel 7.0.12,
AMD Radeon 680M (Rembrandt), libglvnd-1.7.0-9.fc44, webkit2gtk4.1-2.52.5-1.fc44.

netherguy4:
Fixes #1347

## Problem

The AppImage bundles the Ubuntu build of WebKitGTK. On hosts with a newer graphics stack (Mesa/libglvnd on Fedora 42+, atomic distros like Bluefin/Silverblue/Kinoite) that build fails to create an EGL display (`EGL_BAD_PARAMETER`) and `WebKitWebProcess` aborts with SIGABRT on every launch, leaving a permanently blank window. Details, stack trace and environment info in #1347.

None of the usual runtime workarounds help (`WEBKIT_DISABLE_DMABUF_RENDERER=1`, `WEBKIT_DISABLE_COMPOSITING_MODE=1`, `LIBGL_ALWAYS_SOFTWARE=1`, `GDK_BACKEND=x11`) — the incompatibility is at the library level.

## Fix

A CI post-processing step (`scripts/appimage-prefer-host-webkit.sh`) patches the linuxdeploy-generated `AppRun` inside each built AppImage:

- If the host has `libwebkit2gtk-4.1` (detected via `ldconfig`), the app is launched with `LD_LIBRARY_PATH` pointing at the host libdir. `LD_LIBRARY_PATH` outranks the binary's `RUNPATH` (`$ORIGIN/../lib`), so the whole stack resolves from the host — same behaviour users currently get by extracting the AppImage and running the inner binary manually.
- Hosts **without** webkit2gtk-4.1 keep using the bundled stack, unchanged.
- `PSYSONIC_FORCE_BUNDLED_WEBKIT=1` forces the old behaviour as an escape hatch.

Post-processing is the only viable seam in this repo: tauri-bundler generates `AppRun` and packs the squashfs inside a single linuxdeploy invocation, and the `bundle.linux.appimage.files` config only reaches the `usr/` subtree before linuxdeploy runs, so the AppRun can't be customized at build time.

## Verified

On Bluefin (Fedora 44, GNOME Wayland, AMD Radeon 680M, webkit2gtk4.1-2.52.5):

- unpatched 1.50.0 AppImage: `EGL_BAD_PARAMETER`, WebKitWebProcess SIGABRT via coredumpctl, blank window;
- 1.50.0 AppImage patched by this script: renders normally on the host WebKit (`/usr/libexec/webkit2gtk-4.1/WebKitWebProcess`), no crashes;
- `PSYSONIC_FORCE_BUNDLED_WEBKIT=1`: behaves exactly like the unpatched original.

Repacking loses no signatures — the current Linux release assets carry no embedded update info or `.sig`.

## Note

This is a workaround for what is ultimately a tauri-bundler/linuxdeploy limitation. If Tauri's generated AppRun ever learns to prefer the host WebKitGTK itself, this step can simply be dropped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/linux-bundle-test.yml
+++ b/.github/workflows/linux-bundle-test.yml
@@ -72,6 +72,10 @@ jobs:
           VITE_LASTFM_API_SECRET: ${{ secrets.VITE_LASTFM_API_SECRET }}
           APPIMAGE_EXTRACT_AND_RUN: 1
         run: npm run tauri:build -- --bundles ${{ inputs.bundles }} --verbose
+      # Same patch the release channel applies, so a dispatch run exercises it
+      # and the artifact is what users would actually download.
+      - name: patch AppImage to prefer host WebKitGTK
+        run: scripts/appimage-prefer-host-webkit.sh src-tauri/target/release/bundle/appimage
       - name: verify bundle output
         run: |
           set -euo pipefail
--- a/.github/workflows/reusable-channel-publish.yml
+++ b/.github/workflows/reusable-channel-publish.yml
@@ -401,6 +401,16 @@ jobs:
           VITE_LASTFM_API_SECRET: ${{ secrets.VITE_LASTFM_API_SECRET }}
           APPIMAGE_EXTRACT_AND_RUN: 1
         run: npm run tauri:build -- --bundles deb,rpm,appimage
+      - name: patch AppImage to prefer host WebKitGTK
+        env:
+          APPIMAGE_EXTRACT_AND_RUN: 1
+        run: |
+          curl -fsSL -o /tmp/appimagetool \
+            https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage
+          chmod +x /tmp/appimagetool
+          for f in src-tauri/target/release/bundle/appimage/*.AppImage; do
+            APPIMAGETOOL=/tmp/appimagetool scripts/appimage-prefer-host-webkit.sh "$f"
+          done
       - name: upload Linux artifacts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/reusable-channel-publish.yml
+++ b/.github/workflows/reusable-channel-publish.yml
@@ -401,6 +401,8 @@ jobs:
           VITE_LASTFM_API_SECRET: ${{ secrets.VITE_LASTFM_API_SECRET }}
           APPIMAGE_EXTRACT_AND_RUN: 1
         run: npm run tauri:build -- --bundles deb,rpm,appimage
+      - name: patch AppImage to prefer host WebKitGTK
+        run: scripts/appimage-prefer-host-webkit.sh src-tauri/target/release/bundle/appimage
       - name: upload Linux artifacts
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
+### The AppImage starts on Fedora and the atomic distros again
+
+**By [@netherguy4](https://github.com/netherguy4), PR [#1348](https://github.com/Psysonic/psysonic/pull/1348)**
+
+* On Fedora 42+ and the atomic distros built on it (Silverblue, Kinoite, Bluefin) the AppImage opened an empty window and nothing ever appeared in it. The WebKitGTK copy shipped inside the AppImage cannot start on those systems' newer graphics stack, and it crashed on every launch.
+* The AppImage now launches against your system WebKitGTK whenever one is installed, which is the copy that works there. Systems without one keep using the bundled copy exactly as before, and `PSYSONIC_FORCE_BUNDLED_WEBKIT=1` forces the bundled copy back if you ever need it.
+
 ### Shared Top Albums pictures show their covers again
 
 **By [@Psychotoxical](https://github.com/Psychotoxical), PR [#1478](https://github.com/Psysonic/psysonic/pull/1478)**
--- a/README.md
+++ b/README.md
@@ -206,6 +206,8 @@ curl -fsSL https://raw.githubusercontent.com/Psysonic/psysonic/main/scripts/inst
 Linux builds are also available through GitHub Releases, AUR and Cachix/Nix.
 
 > **AppImage runs under X11/XWayland** — it pins `GDK_BACKEND=x11` for a stable WebKitGTK stack. For a native-Wayland launch, use the `.deb`, `.rpm`, AUR, or Nix packages, which follow your session's display server.
+>
+> **AppImage uses your system WebKitGTK when you have one** — if `libwebkit2gtk-4.1` is installed, the AppImage launches against it instead of the copy it ships, which is what makes it work on Fedora and the atomic distros (Silverblue, Kinoite, Bluefin), where the bundled build cannot start. Without it the bundled copy is used, unchanged. Should the system one ever misbehave, `PSYSONIC_FORCE_BUNDLED_WEBKIT=1 ./Psysonic_*.AppImage` goes back to the bundled stack.
 
 ## Windows
 
--- a/scripts/appimage-prefer-host-webkit.sh
+++ b/scripts/appimage-prefer-host-webkit.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# Post-process a Tauri-built AppImage so its AppRun prefers the host WebKitGTK
+# stack when one is available.
+#
+# Rationale: the AppImage bundles the Ubuntu build of WebKitGTK. On hosts with
+# a newer graphics stack (Mesa/libglvnd on Fedora 42+, atomic distros like
+# Bluefin/Silverblue) that build fails to create an EGL display
+# (EGL_BAD_PARAMETER) and WebKitWebProcess aborts, leaving a blank window.
+# The host WebKitGTK, when installed, is always a better fit; the bundled copy
+# remains as a fallback for hosts without webkit2gtk-4.1.
+#
+# Usage: appimage-prefer-host-webkit.sh <path-to.AppImage>
+# Requires: appimagetool (path via $APPIMAGETOOL, default: from $PATH).
+set -euo pipefail
+
+appimage="$(readlink -f "$1")"
+appimagetool="${APPIMAGETOOL:-appimagetool}"
+
+workdir="$(mktemp -d)"
+trap 'rm -rf "$workdir"' EXIT
+cd "$workdir"
+
+"$appimage" --appimage-extract >/dev/null
+
+apprun=squashfs-root/AppRun
+# Sanity check: only patch the linuxdeploy-generated AppRun we know the shape of.
+grep -q 'apprun-hooks' "$apprun" || {
+    echo "error: unexpected AppRun format in $appimage, refusing to patch" >&2
+    exit 1
+}
+grep -q 'PSYSONIC_FORCE_BUNDLED_WEBKIT' "$apprun" && {
+    echo "AppRun already patched, nothing to do"
+    exit 0
+}
+
+binary="$(ls squashfs-root/usr/bin)"
+[ "$(echo "$binary" | wc -l)" -eq 1 ] || {
+    echo "error: expected exactly one binary in usr/bin" >&2
+    exit 1
+}
+
+block="$(cat <<'EOF'
+
+# Prefer the host WebKitGTK stack when available. The bundled (Ubuntu) build of
+# WebKitGTK is incompatible with newer host graphics stacks (Mesa/libglvnd on
+# e.g. Fedora 42+): EGL display creation fails with EGL_BAD_PARAMETER and
+# WebKitWebProcess aborts, leaving a blank window. LD_LIBRARY_PATH outranks the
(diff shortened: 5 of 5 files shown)


Psychotoxical:
Thanks — the diagnosis in #1347 is solid and the direction is right. One blocker, then some smaller things.

**Blocker: the patch execs before the AppRun hooks are sourced.**

linuxdeploy generates:

```sh
this_dir="$(readlink -f "$(dirname "$0")")"
source "$this_dir"/apprun-hooks/...
exec "$this_dir"/AppRun.wrapped "$@"
```

The awk inserts after `this_dir=` and the injected block `exec`s, so on every host that *does* have webkit2gtk-4.1 — the target case — all hooks are skipped, including the one that pins `GDK_BACKEND=x11`. That pin is a documented property of the AppImage channel (README + changelog), and the app only *reads* `GDK_BACKEND` in `src-tauri/src/main.rs`; nothing restores it.

That matters beyond tidiness: in #1072 two reporters hit a permanent `WebKitWebProcess` hang on NVIDIA + Wayland while running the distro packages — already on the **host** WebKitGTK — and `GDK_BACKEND=x11` is what fixes it. Host WebKitGTK alone doesn't make Wayland safe on NVIDIA, and the AppImage is currently the only channel that pins. Your verification was on AMD and necessarily ran without the hooks, so it doesn't cover that combination.

Suggested change: insert the block **before** the final `exec` and drop the `exec` from it — just prepend to `LD_LIBRARY_PATH`. Running after the hooks still puts the host libdir first, keeps everything else intact, and sidesteps whether `usr/bin/<binary>` and `AppRun.wrapped` are equivalent.

**One test that would settle it.** Extract the AppImage and run, with the host libdir `ldconfig -p` reports:

```sh
GDK_BACKEND=x11 LD_LIBRARY_PATH=/usr/lib64 ./squashfs-root/usr/bin/psysonic
```

If that renders, the pin costs nothing and this PR becomes risk-free. If it doesn't, that's a concrete data point — the x11-vs-session decision belongs in `main.rs`, where the NVIDIA quirks already run before GTK init, so it can be made once for all packages instead of implicitly for one.

**Smaller items**

- `appimagetool` is pulled from `continuous` with no checksum and repacks the artifacts users download. Please pin a tagged release and verify the sha256.
- The glob has no `nullglob`: with no AppImage produced, the loop runs on the literal pattern and the publish job fails after a successful build.
- `ldconfig` isn't on a non-root `PATH` everywhere (Debian) — detection then silently falls back to the bundled stack. `PATH=/usr/sbin:/sbin:$PATH ldconfig -p` covers it.
- x86_64 is hardcoded three times (tool URL, `awk` filter, `ARCH=`) — fine today, silently wrong for an aarch64 build.
- `PSYSONIC_FORCE_BUNDLED_WEBKIT=1` is a good escape hatch but undiscoverable, since the failure mode is "app doesn't start" — worth a line in the README next to the existing AppImage note.
- The script only runs in the release publish workflow, so nothing exercises it before a real release. Worth a dry run on a pre-release channel.
- README's AppImage note and a changelog entry need updating once the behaviour is settled.

netherguy4:
Thanks — the blocker is real, and I confirmed it. But the suggested fix doesn't hold, and the reason is worth writing down, because it also explains why the `exec` was there in the first place.

**The `exec` is load-bearing.**

I built the suggested shape — block before the final `exec`, no `exec` of its own, just prepend to `LD_LIBRARY_PATH` — and ran it on the failing host. It still launched the *bundled* WebKit and still died:

```
50253 ././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess 4 24
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

`strings` on `AppRun.wrapped` says why:

```
LD_LIBRARY_PATH=%s/usr/lib/:%s/usr/lib/i386-linux-gnu/:%s/usr/lib/x86_64-linux-gnu/:...:%s
```

It doesn't append to `LD_LIBRARY_PATH`, it rebuilds it with the bundled libdirs in front and the inherited value last. Anything the AppRun exports before it is outranked. So falling through to `AppRun.wrapped` puts the bundled stack back, every time.

**The shape that satisfies both constraints** is your placement plus the original `exec`: insert after the hooks, before the final `exec`, and exec the binary directly.

```sh
source "$this_dir"/apprun-hooks/"linuxdeploy-plugin-gtk.sh"        # GDK_BACKEND=x11 comes from here
source "$this_dir"/apprun-hooks/"linuxdeploy-plugin-gstreamer.sh"

if [ -z "${PSYSONIC_FORCE_BUNDLED_WEBKIT:-}" ]; then
    psysonic_host_webkit="$(PATH=/usr/sbin:/sbin:$PATH ldconfig -p 2>/dev/null | awk '/libwebkit2gtk-4\.1\.so\.0 .*x86-64/{print $NF; exit}')"
    if [ -n "$psysonic_host_webkit" ] && [ -e "$psysonic_host_webkit" ]; then
        export LD_LIBRARY_PATH="$(dirname "$psysonic_host_webkit")${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
        export PATH="$this_dir/usr/bin:$PATH"
        exec "$this_dir"/usr/bin/psysonic "$@"
    fi
fi
exec "$this_dir"/AppRun.wrapped "$@"
```

On the `usr/bin/<binary>` vs `AppRun.wrapped` question: what `AppRun.wrapped` sets beyond `LD_LIBRARY_PATH` is `PATH`, `PYTHONPATH`, `PERLLIB`, `QT_PLUGIN_PATH` and `GST_PLUGIN_SYSTEM_PATH{,_1_0}`. The gstreamer hook already sets `GST_PLUGIN_SYSTEM_PATH_1_0`, the gtk hook already sets `APPDIR` and `XDG_DATA_DIRS`, `PATH` is carried above, and the rest is Qt/Python/Perl we don't ship. The fallback branch still goes through `AppRun.wrapped` untouched.

**Your test, run.** Bluefin (Fedora 44, GNOME Wayland, AMD Radeon 680M, webkit2gtk4.1-2.52.5), host libdir per `ldconfig -p` is `/lib64`:

- `GDK_BACKEND=x11 LD_LIBRARY_PATH=/lib64 ./squashfs-root/usr/bin/psysonic` — renders, host `WebKitWebProcess`, no crash. **The x11 pin costs nothing here**, so keeping it is free on this combination.
- Full patched AppImage, default path — renders, `/usr/libexec/webkit2gtk-4.1/WebKitWebProcess`, and `/proc/<pid>/environ` on the live process shows `GDK_BACKEND=x11`, `LD_LIBRARY_PATH=/lib64:…`, `GTK_THEME`, `APPDIR`, `GST_PLUGIN_SYSTEM_PATH_1_0` — hooks intact, host libdir first.
- `PSYSONIC_FORCE_BUNDLED_WEBKIT=1` — `EGL_BAD_PARAMETER`, exactly #1347.

What this does **not** cover is #1072: NVIDIA + Wayland, which I have no hardware for. This PR doesn't change the x11 pin either way, so it can't regress that case, but it also can't be the thing that proves host WebKitGTK is safe there. If you want the x11-vs-session decision moved into `main.rs` for all packages, that seems right to me, but it's a separate change and I can't validate it.

**Two more bugs I hit while testing this**, both of which would have broken the actual release run:

- `APPIMAGE_EXTRACT_AND_RUN: 1` at step level applies to `"$appimage" --appimage-extract` too, and the AppImage runtime then *runs the app* instead of extracting. `appimagetool` was reached with no `squashfs-root` to pack. The variable is now set only for the `appimagetool` invocation.
- With that variable, `appimagetool`'s own runtime chdirs into its extraction directory, so a relative AppDir path doesn't resolve. It gets an absolute one now.

**Smaller items, all addressed:**

- `appimagetool` pinned to `1.9.1` and verified against a hardcoded sha256 per architecture. (Heads up: the 1.9.1 asset still self-reports `continuous build (git version 8c8c91f)` — that's upstream's labelling, the asset itself is immutable and its hash differs from `continuous`.)
- Architecture derived once from `uname -m`, feeding the tool URL, `ARCH=`, and the `ldconfig -p` filter token (`x86-64` / `AArch64`); an unsupported one fails loudly instead of silently mispatching.
- `ldconfig` looked up through `PATH=/usr/sbin:/sbin:$PATH`.
- `nullglob`, plus a missing bundle directory is skipped rather than passed through as a literal pattern, so a build without an AppImage no longer fails the job.
- README gets a paragraph next to the existing x11 note covering both the host-WebKit behaviour and `PSYSONIC_FORCE_BUNDLED_WEBKIT=1`; CHANGELOG entry under 1.53.0 → Fixed.
- Dry run before a release: `linux-bundle-test.yml` landed since I opened this, so the same step is wired in there. A `workflow_dispatch` run now exercises the script and the artifact is what users would download.

The script also takes a directory now, so both workflows call it as a single line and the appimagetool download, checksum and architecture logic live in one place.

PR #1361, opened 2026-07-28
fix(deps): resolve Dependabot security alerts

cucadmuh:
## Summary
- migrate declarative routing from `react-router-dom` 7.17.0 to `react-router` 8.3.0 and update import sources
- update PostCSS to 8.5.24, `brace-expansion` to 5.0.8, and `quinn-proto` to 0.11.15
- declare Router 8 minimums explicitly: Node 22.22+ and React/React DOM 19.2.7+
- regenerate bundled open-source license data for the resolved dependency graph
- make the boot-chunk and feature-barrel guards resolve repository paths correctly on Windows

## Router 8 decision
Four Router advisories are also fixed by the 7.18 minor line, and GHSA-qwww-vcr4-c8h2 targets RSC mode, which this declarative `BrowserRouter` application does not use. The major upgrade is deliberate rather than required by every alert: Router 8 removes `react-router-dom`, the migration is clean for this application, and adopting it now avoids carrying the compatibility package into the next dependency cycle.

The accepted costs are documented and enforced in the manifest: ESM-only publishing, Node `>=22.22.0`, React/React DOM `>=19.2.7`, and the existing Vite 8 toolchain. `BrowserRouter` and all current hooks correctly import from `react-router`; the app has no `RouterProvider`, `HydratedRouter`, RSC, SSR, or CommonJS router consumer.

## Security
Resolves the seven open Dependabot alerts present on `main` at the time of the update:
- GHSA-qwww-vcr4-c8h2
- GHSA-chx6-hx7r-mcp5
- GHSA-wrjc-x8rr-h8h6
- GHSA-337j-9hxr-rhxg
- GHSA-h8fp-f39c-q6mh
- GHSA-r28c-9q8g-f849
- GHSA-4w2j-m93h-cj5j

`npm audit --audit-level=low` reports zero vulnerabilities after the update.

## Verification
- `npm ci`
- `npm run lint`
- `npm run dep:check`
- `npm test` (503 files, 3560 tests)
- `npx tsc --noEmit`
- `npx vitest run --coverage`
- `npm run build:verify`
- `npm audit --audit-level=low`
- `cargo llvm-cov --workspace` (LCOV + JSON)
- `bash scripts/check-hot-path-coverage.sh`

## Tauri boundary
No commands, events, or payload shapes changed.


Psychotoxical:
Reviewed at `4dd5c162`. Checked out in an isolated worktree with its own `npm ci`: lint, `dep:check`, `tsc --noEmit`, `vitest run` (503 files / 3560 tests) and `build` all pass, `npm audit` reports zero. CI is green across all 12 checks.

Migration itself is clean:

- No `react-router-dom` references remain (109 files carried one on `main`).
- The `react-router/dom` split does not apply — there is no `RouterProvider`, `createBrowserRouter` or `HydratedRouter`, so `BrowserRouter` correctly stays on `react-router`.
- ESM-only is safe: nothing `require()`s the router.
- Baselines hold: Vite 8.0.16, react/react-dom resolve to 19.2.7, and the flake's `nodejs_22` is 22.23.1 in the pinned nixpkgs revision, so the new Node floor does not break the dev shell.

Two notes, neither blocking:

**Scope.** Four of the five router advisories are already fixed in `react-router-dom@7.18.0`. The only one that requires 8.3.0 is GHSA-qwww-vcr4-c8h2, which targets RSC mode — not reachable here, since the app uses declarative routing with no RSC or SSR. Of the 7.18.0 items only the open redirect looks plausibly reachable (pasted share links feed `useNavigate`), and that is covered by the minor too. Merging the major is still reasonable, because `react-router-dom` disappears in v8 either way and this upgrade verifies clean — but the description implies all seven alerts forced it. Worth recording that the major was a deliberate call, along with what it pulls in: the Node 22.22 floor, ESM-only publishing, and the Vite 7 coupling.

**Declared React range.** `package.json` keeps `react: ^19.2.5` / `react-dom: ^19.2.6` while react-router 8 requires `>= 19.2.7`. The lockfile resolves correctly today, so nothing breaks, but a future lockfile rebuild could land below the peer requirement. Raising both to `^19.2.7` would make the constraint honest.

Separately, `src/data/licenses.json` still lists `react-router-dom`, `cookie` and `set-cookie-parser` and lacks `cookie-es`. It was already stale before this branch, so it is not a merge condition, just worth a regeneration pass at some point.

One unrelated find while verifying: `scripts/check-boot-chunk-lucide.mjs:14` builds its path via `new URL('..', import.meta.url).pathname`, which produces an invalid path on Windows, so the guard aborts with "dist/assets not found" after a successful build. Re-run with `fileURLToPath` it reports `ok`. Same class as `scripts/check-feature-barrel-ui.mjs:16` — a small chore PR could fix both.

Approve.

cucadmuh:
Addressed all review notes in `067a4a67`:

- raised declared `react` and `react-dom` ranges to `^19.2.7`
- regenerated `src/data/licenses.json` (now includes `react-router` and `cookie-es`; stale removed npm packages are gone)
- fixed both Windows-unsafe `import.meta.url` path conversions with `fileURLToPath`
- updated the PR description to record the deliberate Router 8 trade-off and its Node/React/ESM/Vite implications

Re-ran the complete frontend CI block locally: 503 files / 3560 tests, lint, dependency-cruiser, TypeScript, coverage and hot-path gate, production `build:verify`, both guard scripts, and `npm audit` all pass.

Psychotoxical (approved):
Re-checked at `3c99909f`. All four notes from the previous pass are addressed.

- **Declared React range** — `react` and `react-dom` are now `^19.2.7`. The lockfile diff touches only the range lines, resolution stays at 19.2.7.
- **`licenses.json`** — regenerated (1.48.0-dev to 1.51.0-dev). I cross-checked every npm entry against the production closure of the lockfile: 74 entries against 74 packages, nothing missing, nothing stale, no version drift. One correction to my earlier note: `cookie` is still listed, but that is the cargo crate 0.18.1, which belongs there. The npm package `cookie@1.1.1` is the one that is gone, along with `react-router-dom` and `set-cookie-parser`, with `cookie-es` added.
- **Router 8 rationale** — now on record in the description, including the 7.18 alternative, the RSC scope of GHSA-qwww-vcr4-c8h2, and the Node/React/ESM costs.
- **Windows path traps** — both guards converted to `fileURLToPath`.

I verified the guards by running them against each other on Windows rather than reading the diff. Same input, `main` version versus this branch:

| Guard | `main` | this branch |
|---|---|---|
| `check-feature-barrel-ui.mjs` | false positive, `missing barrel file`, exit 1 | `ok`, exit 0 |
| `check-boot-chunk-lucide.mjs`, with `dist/assets` present | `dist/assets not found`, exit 1 | `ok`, exit 0 |

A seeded `createLucideIcon` chunk still trips the boot-chunk guard, so it reads the chunks rather than merely locating the directory. This also unblocks `dep:check:barrels` on Windows, which had been reporting phantom violations there.

CI is green across all 12 checks on this head, including both Windows jobs.

Still outside every automated gate: a manual pass through navigation, back button, scroll restore and share-link paste in a running dev build. A router major is exactly the class of change the suite does not exercise in the webview.

Approve.

PR #1363, opened 2026-07-29
fix(playback): keep Hot Cache files scoped to queue servers

cucadmuh:
## Summary

- resolve each Hot Cache prefetch job from the upcoming queue item's canonical server instead of applying the current playback server to the whole window
- revalidate pending jobs with the full `serverId:trackId` identity so a stale job cannot survive a queue switch to the same track ID on another server
- key eviction protection and queue tier lookup by canonical `serverId:trackId`, preserving current/next tracks from every server before inactive-server eviction
- calculate protected-window cache usage with each queue item's own server scope
- add regression coverage for mixed-server prefetch, delayed stale-job replacement, and over-budget eviction with reused track IDs

## User impact

Hot Cache downloads now stay with the server that owns each queued track, including mixed-server queues, queue replacements where servers reuse track IDs, and full-cache eviction passes.

## Verification

- `npm run lint`
- `npm run dep:check`
- `npm test` (505 files, 3563 tests)
- `npx tsc --noEmit`
- `npx vitest run --coverage`
- `bash scripts/check-frontend-hot-path-coverage.sh` (15/15 files at or above 70%)

## Tauri boundary

No commands, events, or payload shapes changed. The existing `download_track_local` call now receives the correct per-item server scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -219,6 +219,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * On large libraries the Artists page could show nothing but a spinner, indefinitely. It now loads in well under a second.
 * Root cause: the query that collects the artists in the selected music folders was combining two steps in the wrong order, redoing the expensive one once per artist instead of once in total. On a library of this size that meant the query never finished at all.
 
+### Hot Cache — prefetch mixed-server queues from the right server
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#1363](https://github.com/Psychotoxical/psysonic/pull/1363)**
+
+* Hot Cache now downloads each upcoming track from its owning server instead of sending the whole prefetch window to whichever server is currently playing. Replacing the queue can no longer leave a delayed job caching the wrong copy when two servers reuse the same track ID.
+
 
 ## [1.50.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,6 +224,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1363](https://github.com/Psychotoxical/psysonic/pull/1363)**
 
 * Hot Cache now downloads each upcoming track from its owning server instead of sending the whole prefetch window to whichever server is currently playing. Replacing the queue can no longer leave a delayed job caching the wrong copy when two servers reuse the same track ID.
+* When the cache is full, the current and next tracks stay protected per server, so an identically numbered track on another server cannot evict the queued copy and trigger a download loop.
 
 ### AppImage — installing through an AppImage manager
 
--- a/src/hotCachePrefetch.test.ts
+++ b/src/hotCachePrefetch.test.ts
@@ -1,0 +1,106 @@
+import { waitFor } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { initHotCachePrefetch } from '@/hotCachePrefetch';
+import { usePlayerStore } from '@/features/playback/store/playerStore';
+import { useLocalPlaybackStore } from '@/store/localPlaybackStore';
+import { useAuthStore } from '@/store/authStore';
+import { setDeferHotCachePrefetch } from '@/lib/cache/hotCacheGate';
+import { makeServer, makeTrack, seedQueue } from '@/test/helpers/factories';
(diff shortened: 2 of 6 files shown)


Psychotoxical:
Reviewed at `71e84a13`. Local checks on the touched paths: `lint`, `dep:check`, `tsc --noEmit` and `vitest run` (504 files / 3562 tests) all pass. Coverage and the hot-path script I did not run locally and took from your report.

The change reads right, and two things beyond the stated goal are worth noting: `index === 0` also fixes a duplicate-track case the old `trackId` comparison got wrong (the same track twice in a queue matched the immediate-next test from any position), and the redundant second `jobRef` lookup is gone.

One follow-up that deserves a conscious decision rather than being inherited — **eviction still scopes to a single server.**

`replanNow` calls `evictToFit(..., playbackSid, ...)` before planning anything, and inside `evictEphemeralToFit` (`src/store/localPlaybackStore.ts`) `protectedIds` holds bare track IDs, an entry survives only when `protectedIds.has(trackId) && serverIndexKey === activeServerIndexKey`, and every other server's entry is pushed as `tier: 0` — the first class deleted, since candidates sort ascending by tier.

So in an over-budget mixed-server queue, the track this PR now correctly prefetches for the other server is the top eviction candidate even while it sits inside the protected window. The next `replanNow` re-queues it, eviction drops it again.

It is pre-existing and bounded — the function returns early while `diskBytes <= maxBytes` — but it was effectively unreachable before this PR, since nothing produced foreign-server prefetches on that path. Keying `protectedIds` by `serverIndexKey:trackId` and gating the `tier: 0` branch on "outside the protected window" rather than on server identity would close it. Your call whether that belongs here or in a follow-up; the reason for raising it is that merging this alone leaves the mixed-server case half-correct once the cache fills.

Minor: the `skipped.push({ trackId, reason })` entries in `replanNow` carry no server, while the job-level skip log gained one in this diff — with two servers sharing track IDs those skip reasons can no longer be attributed.

Separately, the PR is DIRTY at the moment: one commit behind `main`, and the changelog addition conflicts now that #1364 and #1365 have appended to the same section.

cucadmuh:
Addressed the review points in `d292239d` and synced the branch with current `main` (`48d80ab2`).

- Eviction protection now uses the canonical composite `serverIndexKey:trackId` for every current/next queue ref, so a protected foreign-server entry bypasses the inactive-server tier.
- Queue membership/tier lookup uses the same composite key, so reused track IDs on another server no longer make an unrelated entry look queued.
- Added `src/store/localPlaybackStore.test.ts`: under budget pressure with A/B sharing `trackId`, the unused A copy is evicted and the protected B copy remains.
- `replan` skip diagnostics now include `serverId`.
- Resolved the changelog conflict in PR order (`#1363` before `#1364`) and expanded the entry for full-cache behaviour.

Local verification is green: `lint`, `dep:check`, `tsc --noEmit`, `npm test` (505 files / 3563 tests), release-note generation, coverage, and the 15/15 hot-path gate. GitHub CI is rerunning on the new head.

Psychotoxical (approved):
Re-reviewed at `d292239d`. Both points are closed, and I checked each rather than taking the summary on trust.

The eviction fix is at the right place: `protectedKeys` now holds the composite `serverIndexKey:trackId`, and the check sits **before** the inactive-server branch — which is why `tier: 0` could stay as it was. A protected foreign-server entry no longer reaches it, so the "entries of a server we are not playing are cheapest to drop" heuristic survives, just scoped to outside the protected window. That is the right scope for it.

The `indexOfInQueue` change closes a second defect of the same class that I had not named: membership was resolved with `findIndex` over bare track IDs, so an entry owned by one server could be treated as queued at a position actually held by another. The composite-key map also turns a per-candidate scan into a single pass.

Key derivation stays consistent with the prefetch side — `canonicalQueueServerKey` never returns empty for a non-empty input, so the `|| activeServerIndexKey` fallback is reachable only for an empty `serverId`, and there both paths land on the same value since `replanNow` passes `getPlaybackCacheServerKey()` as the scope.

`localPlaybackStore.test.ts` is a real regression test, not a passing one: against the pre-fix code the outcome inverts — `a.test:shared` would have been protected via the bare track ID and `b.test:shared` would have been the first `tier: 0` deletion. `LOCAL_PLAYBACK_PROTECT_AFTER_CURRENT` is `1`, so the asserted index really is inside the protected window.

Changelog order is right (the #1363 block precedes #1364), and the branch is CLEAN with all six checks green.

One transparency note: I did not re-run the local frontend block on this head — the clone was on an unrelated branch in active use, and I did not want to disturb it. This pass is static review plus the green CI run.

Approving.

PR #1365, opened 2026-07-29
fix(library): let server-side deletions leave the index

Psychotoxical:
Server-side deletions stopped reaching the index because the forward-only delta could not observe them and the old threshold gate no longer accumulated enough stale rows to run. This PR now reconciles that state without treating partial visibility or an incomplete ingest as proof that music disappeared.

## What changed

- Added a bounded background album census that compares server and local inventories, confirms removals with `getAlbum`, batches tombstones and projection cleanup, and fills albums missed below the delta watermark.
- Hardened full-resync pagination and destructive sweep authorization. Server page clamps no longer skip ranges; a sweep requires an exact, fresh, non-scanning server count.
- Split sparse and authoritative merge behavior while keeping field-level observation semantics: `sortTitle`, `orderTitle`, OpenSubsonic `sortName`, and `updatedAt` only clear when explicitly observed as null. `getAlbum` remains authoritative for album credits.
- Preserved raw OpenSubsonic album/song extensions, library scope, genre projection, sort titles, and delta watermarks across partial payloads.
- Batched confirmed album retirement into one transaction and aligned identity handling for guest credits, compilations, and collection labels.
- Fixed server-profile removal so the real URL-keyed session and rows are cleared, while profiles sharing one normalized URL keep and rebind the shared index.
- Fixed API-seeded album/offline rows using epoch seconds for `synced_at`; they now use the same millisecond clock as every sync path.

## Live verification

The reworked sync was exercised on two Navidrome servers, including a 175,468-row `search3` resync and server-side album deletion. Counts matched, existing sort titles/watermarks/credits survived, confirmed removals tombstoned only the expected tracks, and projection plus genre cleanup completed together. Removing the final server profile also purged its URL-keyed state cleanly.

Navidrome continues to advertise deleted files while they remain under **Missing Files**. Psysonic can retire an album only after those entries are cleared and the server stops returning it.

The focused recheck on `84e885f0` confirmed both follow-up fixes against the live databases. Opening one eight-track album wrote a thirteen-digit millisecond `synced_at` and populated `title_sort` on exactly those eight rows from OpenSubsonic `sortName`. A forced census gap genuinely rewrote four rows while preserving their existing `server_updated_at` watermark and sort titles; an explicit JSON null still cleared the field as intended.

Not reproduced live: a server that clamps page size, scoped full resync, and two profiles sharing one normalized URL. Deterministic tests cover those branches and the scoped path deliberately fails closed.

## Verification

- `cargo test --manifest-path src-tauri/Cargo.toml --workspace --all-targets`: pass, including 768 `psysonic-library` tests and 13 ignored probes.
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings`: pass.
- `cargo llvm-cov --manifest-path src-tauri/Cargo.toml --workspace --json --output-path src-tauri/target/llvm-cov/cov.json`: pass.
- `bash scripts/check-hot-path-coverage.sh`: all 11 files above 70%.
- `npm run prebuild:release-notes`: pass.
- Full frontend CI-equivalent gates passed on the preceding reworked head; every GitHub Actions check for `84e885f0`, including `ci-ok`, passed.

## Contracts

No Tauri command name, argument shape, return payload, event name, or event payload changed. No database migration or on-disk schema change is required.

No blocking review finding remains. The PR is ready for a human maintainer merge decision.


cucadmuh:
## Review update: full verification required before merge

The reviewed fixes are now pushed at `405fb5c5e24addd7d2fc96b2c2b1f7136964360c`, including the merge from current `origin/main`. The complete frontend and Rust CI-equivalent suites passed locally after that merge, including lint, dependency-cruiser, Vitest, TypeScript, frontend coverage/hot-path gates, workspace Rust tests, Clippy, and Rust coverage/hot-path gates.

**Do not merge this PR until both conditions below are satisfied and recorded here:**

1. All newly triggered GitHub Actions finish successfully, including the aggregate `ci-ok` check.
2. The following live-server scenarios are exercised against backed-up test databases:

- Run a full S1/search3 resync on a large server and compare server, stamped, and live track counts; confirm no contiguous range is skipped and existing credits, sort titles, raw extensions, and update watermarks remain intact.
- Repeat against a server that clamps requested page sizes; confirm search and album-list walks continue safely to an empty page without skipping or looping.
- Start a server scan while a scheduler tick is due; confirm census and library tagging are deferred, no rows are removed, and retry occurs after roughly 30 seconds.
- Remove one album from a library with fewer than ten albums; after the server rescan, confirm retirement occurs only after direct `getAlbum` NotFound confirmation.
- Remove one album from a large library; confirm album/composer surfaces and `local_track_count` update correctly through the batched removal path.
- Add an album below the delta watermark; confirm census gap ingestion preserves raw album/song extensions and refreshes the UI.
- Exercise playback while census is due on the largest available library; confirm census yields, completes within budget, or reports `budget_exhausted` without false refreshes or one-minute retry churn.
- Run scoped and no-scan-count full resyncs; confirm stale rows are retained with a skip diagnostic, then verify genuinely dead tracks through the integrity workflow.
- Remove the final profile for a server; confirm its session stops and optional purge removes URL-keyed rows. Repeat with two profiles sharing one normalized URL and confirm removing one preserves rows and rebinds the session to the surviving credentials.

Please record the server types, approximate library sizes, and pass/fail result for each applicable scenario in this PR. Final merge remains a human maintainer decision.

cucadmuh:
Automated verification update: all checks for `405fb5c5` are now green, including `ci-ok`, frontend lint/dependency checks/tests/coverage, and Linux/Windows Rust tests, Clippy, and coverage.

The PR should still remain unmerged until the live-server checklist in the preceding comment is completed and its results are recorded.

Psychotoxical:
Live-verified the reworked head on two Navidrome servers. Full measurements are in the workdocs handoff `2026-07-30-pr-1365-live-verification.md`.

Core behaviour works: a server-side album deletion leaves the index within one census interval, with per-album confirmation, tombstones rather than hard deletes, and projection plus genre cleanup in one transaction. The large-catalogue `search3` resync preserved sort titles, watermarks and album credits unchanged across 175,468 rows, and removing a server profile purged cleanly with nothing left behind.

Two findings:

**1. The census gap-fill clears `title_sort` and `server_updated_at`.** The sparse/authoritative split is decided per endpoint, but field coverage varies per field within one endpoint: `getAlbum` is authoritative for album credits, yet carries neither `sortTitle`/`orderTitle` nor `updatedAt`. On the live index, of 305,807 rows carrying the album-to-track copy, zero hold either key, while 47,736 hold a sort title and 47,748 a watermark. Reproduced live: the census filled two album gaps and both landed with 0 of 6 sort titles and 0 of 6 watermarks, while credits survived through the display-credit fallback. Bounded in practice — the delta normally wins the race, and the gap-fill fired once across the whole session.

**2. Pre-existing, not from this branch:** `library_upsert_songs_from_api` (`commands.rs:582`) builds `synced_at` with `as_secs()` where every sync path uses milliseconds, so affected rows are off by a factor of 1000. It matters because `synced_at` orders tombstone candidates ascending. It only surfaced here because opening an album detail page invokes that command.

Operational note worth capturing somewhere: the census can only retire an album once the server stops advertising it. Navidrome deliberately keeps deleted files as Missing Files and continues to serve them through the API, so both delete tests produced `removed>0` only after those entries were cleared server-side.

cucadmuh:
Addressed both live-verification findings in `84e885f0`.

- `getAlbum` remains authoritative for album credits, but sort and watermark fields now use field-level presence semantics: omission preserves the stored value and an explicit null clears it. The mapper also now reads OpenSubsonic `sortName`, which Navidrome emits for Subsonic song children, so fresh census gaps receive the available sort key instead of looking only for native `sortTitle` / `orderTitle`. Navidrome does not expose native `updatedAt` through this endpoint, so a fresh gap does not invent a false watermark; existing watermarks are no longer eroded.
- `library_upsert_songs_from_api` now goes through the shared millisecond clock and has a regression test around the stored timestamp.
- The changelog now records the Navidrome **Missing Files** precondition.

Local verification is green: workspace tests, Clippy, Rust coverage and all hot-path gates. A small live recheck of one census-filled album would still be useful to confirm `sortName` on the deployed Navidrome versions before merge.

cucadmuh:
Follow-up head `84e885f0` is green. All GitHub checks passed, including Linux and Windows Rust tests/clippy, Rust and frontend coverage gates, frontend lint/type/tests, dependency-cruiser, and `ci-ok`.

The two live-review findings are fixed and covered by regression tests. A short targeted live recheck of a fresh census gap (`sortName` plus omitted `updatedAt`) and the album-detail `synced_at` unit remains recommended before the human merge decision.

Psychotoxical:
Both live findings are confirmed fixed on `84e885f0`, rechecked against the live databases. Backups of both databases were taken and verified first.

**Album detail — millisecond stamp and `sortName`.** One open of an eight-track album whose rows carried `sortName` but no sort title: `synced_at` moved from `1785445377379` to `1785491917102` — thirteen digits, landing 19.7 s before the read, so the command genuinely ran rather than leaving the rows alone. `title_sort` filled on all eight rows from OpenSubsonic `sortName`, and the live-wide count moved 45,134 → 45,142, exactly eight, so nothing outside the album was rewritten. Rows with a seconds-scale stamp: zero before and after.

Why this is observable at all: the deployed server emits `sortName` on 42,395 live rows and `sortTitle` on none, and 105,615 live rows carry a `sortName` with no sort title.

**Census gap-fill — watermark preservation.** The gap was forced without retiring any track row: only the album's `album_browse_projection` row was removed, which is what `local_album_inventory` reads, so all four track rows stayed live and the offline-tier reconciler had no reason to act. The target carried a watermark and sort titles but no `updatedAt` anywhere in `raw_json`.

```
census: server_albums=11851 local_albums=11850 removed=0 filled=1 stale=0 deferred=0
```

`server_updated_at` stayed at `1782833732000` on all four rows and the sort titles were unchanged, while `synced_at` advanced from `1785445404216` to `1785492436746`. That advance is the load-bearing part: the rows were genuinely rewritten and preservation still held, which "nothing changed" alone would not have shown. The projection row came back field for field.

**Counter-proof.** The same inputs through both versions of the expression — authoritative upsert, no `updatedAt` in the payload, incoming value null, existing value present, sparse flag `0` as bound by `upsert_batch_with_remap`:

```
previous  -> NULL
current   -> 1782833732000
current, explicit JSON null -> NULL
```

So the rule neither loses an existing watermark nor prevents a server from clearing one deliberately.

Library state afterwards is fully restored: the album is back, its projection row matches the pre-test values, and the census schedule returned to its normal interval. No track row was ever marked deleted and no media file was touched.

Still not covered live: a server that clamps page size, scoped resync, and two profiles sharing one normalized URL. Full detail is in the handoff.

cucadmuh:
The focused live recheck satisfies the remaining review recommendation on `84e885f0`. It demonstrates that the affected rows were genuinely rewritten while preserving the watermark, confirms `sortName` ingestion and millisecond `synced_at`, and verifies that explicit null still clears deliberately.

All automated checks remain green, including `ci-ok`. The page-clamp, scoped-resync, and shared-normalized-URL cases remain environment-bound but have deterministic coverage; scoped resync also fails closed by design.

No blocking review finding remains. This PR is ready for a human maintainer merge decision.

cucadmuh (dismissed):
lgtm

PR #1369, opened 2026-07-31
fix(playback): keep the queue in step when a paused stream restarts

### LINKED ISSUE #1368: [Bug]: Streamed audio and displayed song out of sync.
### What's wrong?

Psysonic 1.50.0 Windows 10 x86_64. Remote server Navidrome 0.63.2 Docker on Alpine Linux.
While listening remotely to a playlist, after pausing or stopping the audio for some time (undetermined as of yet), the currently shown Now Playing track and Now Playing screen are not in sync with the currently playing audio. Sometimes Now Playing and the listed currently playing track are a few songs ahead of the audio in the playlist.
Going back a few tracks by clicking on the previous track button will eventually correct this behavior.
The audio time display also shows a flat waveform rather than volume amplitude.
This has happened to me several times in the last couple weeks, and I started noticing it on the previous release, 1.49.0.
Bear with me. My first ever bug report on github.

### Steps to reproduce

1. Create a playlist through Build a Mix.
2. Pause the playback for several minutes. Or stop.
3. Click the play button. 

### What should happen?

Streamed audio should match what is shown as playing.

### What actually happens?

Displayed track is usually ahead by 3-4 tracks in the playlist.

### Psysonic version

1.50.0

### Operating system

Windows

### Install source

.msi installer (Windows)

Navidrome

### Server version

0.63.2

### Logs

```shell

```

### Screenshots / video

_No response_

### Anything else

_No response_

Psychotoxical:
Closes #1368.

After the output stream is released on idle, resuming rebuilds it from zero while the pre-pause position is still on record. The gapless reconciler reads that as a decoder boundary — position under 8 s where it was well above — and advances the queue, on every resume. The display then runs several tracks ahead of the audio, and the waveform belongs to the displayed track rather than the audible one.

* A (re)started stream now clears the position history. It has no past to regress from.
* Gapless transitions are unaffected: they arrive as `track_switched`, never through `audio:playing`, so the reconciler still catches a decoder switch the engine did not announce.

## Test plan

* Regression test in `gaplessQueueAdvance.test.ts` fails on the unfixed code with `expected 't2' to be 't1'` — the reported symptom — and passes with the fix. The four existing reconciler tests stay green.
* Live on Windows against a remote server: play, pause, wait past the idle release, resume. No reconcile fires, the seek restores the pre-pause position, and the display matches what is playing. A natural gapless transition still advances correctly.
* `npx tsc --noEmit`, `npm run lint`, full `vitest run` (505 files, 3572 tests).

## Not addressed

The seek guard clears itself after 5 s and then lets progress handling continue into the reconciler (`audioEventHandlers.ts`). That is a second way stale position history can reach it, but there is no evidence it fired in the reported case, so it is left alone here rather than changed on suspicion.

#1224 is a different symptom — the bar lagging behind the audio — and is not addressed by this change.


cucadmuh (approved):
lgtm

PR #1371, opened 2026-08-01
feat(playback): clickable artists in fullscreen, mini, and mobile-layout player views

enncoded:
## What changed

Artist credits are now clickable in **every player view**, not just album/queue rows:

- **Fullscreen player** (all three styles: minimal, immersive, prism) — click an artist to jump to their page; the overlay closes so the artist page is immediately visible
- **Detached mini player** — artist names link to the artist page in the main window
- **Mobile-layout player** — same behavior

All five player views share one adapter, `TrackArtistLinks`, built on the `resolveTrackArtistRefs` pipeline from #1342:

- Legacy Navidrome credits like `Primary feat. Guest` render as separate artist labels, with guest links resolved from the local library index
- Structured OpenSubsonic `artists[]` remains authoritative (no resolution call when present)
- Also splits `Alice • Bob` — Navidrome's default `Scanner.ArtistJoiner` display name for only-plural `ARTISTS` tags
- `AC/DC` and comma-joined credits stay one label, per the Navidrome separator contract

## Who should notice

End users: artist names in the fullscreen, mini, and mobile players are now links to the artist page.

## How to verify manually

1. Play a track tagged `Primary feat. Guest` (both artists indexed)
2. In each fullscreen style, click `Guest` → the fullscreen overlay closes and the Guest artist page opens
3. In the detached mini player, click a guest artist → the main window navigates to their page
4. With a structured two-artist `artists[]` track, both names link immediately (no name-resolution call)
5. `AC/DC` and comma-joined credits render as one label — never split

## Tauri boundary

Not touched — no commands, events, or payload changes; frontend-only.

## Server compatibility

Navidrome-first, as usual. The `•` separator matches Navidrome's default `Scanner.ArtistJoiner`; servers that only send the legacy `artist` string still work via the existing separator set.

## Tests

Added/updated unit + component tests (`TrackArtistLinks`, `MiniMeta`, all three fullscreen players, `miniPlayerHelpers`, `openArtistRefs`). Local run: `npm test`, `npm run lint`, `npm run dep:check`, `npx tsc --noEmit`, coverage + hot-path gate all green.

## Assumptions for review

Implementation proceeded on these assumptions (from the upstream data-format analysis); happy to adjust if any are wrong:

1. **Track-performer credits (`artists`) are the right field for the player views**, not album-artist credits (`albumArtists`). The album header shows album artists; the player views show the playing track's performers. If you'd prefer album-artist credits in any view, that first needs `albumArtists` mapped into the `Track` type (separate change).
2. **`track.artist` (the legacy string) is the correct split source**, not `displayArtist`. Psysonic currently drops `displayArtist` in `songToTrack`; on Navidrome both values are typically identical. If you'd prefer preferring `displayArtist` when available, that also needs mapping into `Track` first.
3. **The separator contract matches Navidrome's current behavior**, verified against the tagging docs (last modified 2026-07-22). If Navidrome changes its separator list, the regex needs updating — not a blocker, just flagging.
4. **Reusing `resolveTrackArtistRefs` unchanged** (from #1342) seems to be the intended approach for these player views.


Psychotoxical:
Thanks for this — tested it against a real library and it does what it says. Checks are green here too (tsc, lint, dep:check, full vitest run).

I've pushed a changelog entry to your branch, and you're in the in-app credits under Settings → System as of 1.51.0. First contribution to Psysonic — welcome.

Testing it did surface a bug, but not one of yours. When a track carries a structured `artists[]` list, those ids go straight through as link targets without checking whether the local index actually knows them. Track-level performers usually aren't in the artist index at all — it holds album artists — so clicking one of those names leads somewhere it shouldn't.

That comes from the `resolveTrackArtistRefs` pipeline in #1342, not from this PR. `MiniMeta` and the mobile player already had the same path on main. What your change does is put it in front of the fullscreen players, where the artist line wasn't clickable before, which is why it finally showed up. The split path itself handles this correctly — it only links names it resolved against the index — so the fix belongs in the structured branch, and it'll land as a follow-up that covers all five surfaces at once.

Nothing to do on your side.

Issue #1276, opened 2026-07-12
[Bug]: EQ settings are not applied to Internet Radio


mksib:
### What's wrong?

Basically what it says in the title. Trying fiddling with EQ settings while internet radio plays gives zero change in sound. 

### Steps to reproduce

1. Turn on any Internet Radio station
2. Change EQ presets, while station playing
3. Observe absence of change in sound

### What should happen?

EQ should apply to Internet Radio stream as well

### What actually happens?

Nothing

### Psysonic version

1.49.0

### Operating system

Windows

### Install source

.msi installer (Windows)

### Subsonic-compatible server

None

### Logs

```shell

```

PR #1375, opened 2026-08-02
feat(visualizer): Winamp-style audio visualizer

Manwe-777:
## What this adds

A frequency visualizer driven by the audio actually being played, with four modes:

| Mode | What it shows |
|---|---|
| **Spectrum** | Classic log-spaced bars with falling peak caps, over a shared vertical gradient |
| **Oscilloscope** | Time-domain trace with CRT phosphor persistence |
| **Radial scope** | Waveform wrapped around a ring, over a feedback trail that expands and fades outward |
| **Stereo field** | One arc per channel, anchored to its own edge and growing inward, so the stereo image itself is the picture |

Screenshot 2026-08-02 at 1 32 39 AM

Screenshot 2026-08-02 at 1 33 20 AM

**Who should notice:** end users. It is on by default as a card on Now Playing and a strip in both fullscreen players, and can be turned off entirely in Settings → Appearance → Visualizer.

## Why it is built this way

Playback decodes in Rust (`rodio` + `symphonia`), so the webview never sees PCM and the usual `AnalyserNode` approach is unavailable. The design notes below are the non-obvious parts:

**Audio-thread safety.** A `SpectrumTapSource` is inserted into the rodio chain after the EQ and fade stages but *before* the sink — so the visualizer reflects the audio as shaped by the EQ (what you hear) without collapsing to nothing when the volume is lowered. It mirrors samples into a lock-free ring of `AtomicU32` (`f32::to_bits`, so the whole path stays safe Rust — no `UnsafeCell`, no `unsafe impl Sync`). No locks, no allocation, no syscalls on the audio thread; when nobody is watching it costs one relaxed atomic load per sample.

**Crossfade correctness.** During a crossfade two sources are pulled concurrently. A lease guards the ring: each source claims it once, on the first sample it actually *produces* — not at construction, because gapless preload builds the next source long before it is audible. The loser goes quiet rather than interleaving into the same ring.

**Costs nothing when unused.** The FFT task only exists while a visualizer surface is mounted (`audio_spectrum_set_active`), and stops emitting entirely once the envelopes settle after playback ends. The frontend refcounts across surfaces and only crosses the boundary on the 0↔1 edges.

**IPC.** Frames are quantised to bytes and base64'd rather than sent as JSON number arrays — roughly 4× less text per frame, which matters on WebView2 where the pipe is the bottleneck (per the existing note in `ipc.rs`).

**Stereo without a second FFT.** Left and right are kept separate through the ring, so the stereo mode and oscilloscope get real channel separation. The spectrum sums them back to mono on the *reader* side, off the audio thread — a scope is pure time domain, so the split costs one extra relaxed store per frame and no extra transform.

**Colour.** Pulled from the cover art (dominant hue, a distinct second hue, and the brightest pixel) or the theme's accent ramp, then contrast-adapted against the theme background in either direction so the same palette works on light and dark themes. Greyscale artwork falls back to the theme rather than inventing a hue.

## Tauri boundary

Additive only, no renames or shape changes to anything existing:

- **New command** — `audio_spectrum_set_active(active: bool, fps: Option<u32>, responsiveness: Option<f32>)`. Idempotent by design: a duplicate `true` (e.g. a dev reload that lost the frontend's refcount) replaces the running task rather than stacking a second one.
- **New event** — `audio:spectrum`, carrying base64 band/peak/waveform arrays plus `rms`, `peak`, `bandCount`, `waveCount`, `sampleRate`. The frontend sizes its buffers from the payload's counts, so the Rust-side `BAND_COUNT` / `WAVE_COUNT` constants are the single source of truth and can be tuned without a frontend change.
- `src/generated/bindings.ts` regenerated via the repo's `regenerate_committed_bindings` path; `committed_bindings_are_fresh` passes.
- Verified on both sides: Rust unit tests for the tap, FFT, band mapping and payload encoding; Vitest for the feed hook, decoding and renderers.

## Persisted settings

- **New store key** `psysonic_visualizer` (mode, sensitivity, responsiveness, frame rate, peak caps, colour source). Fresh installs get defaults; nothing existing is read or migrated.
- **`psysonic_np_layout`** gains a `visualizer` card. The store's existing `onRehydrateStorage` already appends unknown-but-default cards, so existing installs pick it up at the end of the left column with no data loss and no migration step.

## Server compatibility

No server dependency — this reads the local audio pipeline only. Works identically on Navidrome, Gonic, Airsonic, LMS.

**One limitation worth calling out:** internet radio plays through an `HTMLAudioElement`, so it is read from a Web Audio `AnalyserNode` instead of the Rust tap. That analyser is only available while the equalizer is enabled — attaching one otherwise would require `createMediaElementSource`, which permanently hijacks element output and goes silent on streams without CORS (the exact thing #1276 avoids). I did not want to change that policy in this PR. The limitation is stated in the settings panel.

## i18n

English (`src/locales/en/visualizer.ts`) added and wired into `en/index.ts`. Other locales left for follow-up translation PRs per CONTRIBUTING; `fallbackLng: 'en'` covers them meanwhile.

## Accessibility

Every mode honours `prefers-reduced-motion`: the radial and oscilloscope drop their persistence trails (that persistence is precisely what the preference asks us to suppress), and glow/bloom work is skipped. The canvas is `aria-hidden` — it is decorative, and the audio it depicts is already announced by the player.

## How to verify manually

1. Play a track. The visualizer appears on Now Playing (left column) and in the fullscreen player.
2. Cycle modes with the button on the panel; expand with the arrows — it fills the window down to the player bar, Escape collapses.
3. Settings → Appearance → Visualizer: try each mode, the sensitivity and responsiveness sliders, peak caps, and Album art vs Theme colouring.
4. Toggle it off and confirm the panel disappears everywhere.
5. Switch themes (including a light one such as Catppuccin Latte) and confirm the colours stay legible.
6. Drag the card in the Now Playing layout editor, and hide it — it behaves like every other card.

## Checks run

`npm run lint`, `npx tsc --noEmit`, `npm run dep:check` (no new layering violations), `npm test`, `npm run build`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`. Rebased onto current `main` before opening.


cucadmuh (approved):
lgtm

Issue #1373, opened 2026-08-01
[Bug]: Small gap between tracks in "gapless" mode


smirnoffjr:
### What's wrong?

Gapless mode greatly reduces silence between tracks, but still you can hear small noticeable gap or "hiccup" when next track starting.

### Steps to reproduce

1. Enable gapless mode
2. Play album which can be listened gapless (e.g. Pink Floyd - The Dark Side Of The Moon)
3. Listen carefully during track switching

### What should happen?

Tracks should be played seamlessly

### What actually happens?

You can hear small gap

### Psysonic version

1.50.0

### Operating system

Linux

### Install source

.rpm (Linux)

### Subsonic-compatible server

Navidrome

### Server version

Navidrome 0.63.2

### Logs

```shell
[2026-08-01 20:25:43.296 +0300] [psysonic] logind PrepareForSleep listener registered (post-sleep audio recovery)
[2026-08-01 20:25:44.601 +0300] [psysonic] audio stream opened at 44100 Hz (exact)
[2026-08-01 20:26:13.736 +0300] [library-db] write op=track.upsert_batch_remap lock_wait_ms=0 exec_ms=206
[2026-08-01 20:26:13.737 +0300] [library-db] write op=fact.get_gc lock_wait_ms=206 exec_ms=0
[2026-08-01 20:29:58.423 +0300] [psysonic] audio output stream released after 60s idle
[2026-08-01 20:31:02.645 +0300] [psysonic] audio stream opened at 44100 Hz (exact)
[2026-08-01 20:37:38.557 +0300] [psysonic] audio output stream released after 60s idle
[2026-08-01 21:17:13.115 +0300] [psysonic] audio stream opened at 44100 Hz (exact)
[2026-08-01 21:26:04.259 +0300] [psysonic] audio output stream released after 60s idle
[2026-08-01 21:27:11.521 +0300] [psysonic] audio stream opened at 44100 Hz (exact)
[2026-08-01 21:27:43.368 +0300] [analysis] queue_snapshot interval_s=60 note=queues_in_memory_cleared_on_app_restart | http_backfill={queued:0 tiers=(0,0,0) download_active:0} | cpu_seed={queued_jobs:0 tiers=(0,0,0) decoding_active:0}
[2026-08-01 21:27:43.996 +0300] [analysis] backfill enqueue (analysis pending) track_id=RyTWKGjEUtDjMMjzMBrNPj
[...]

PR #1388, opened 2026-08-05
fix(audio): trim MP3 encoder delay and padding during gapless playback

Psychotoxical:
## Summary

- MP3 encoder delay and padding were played as audio. Symphonia reads both from the Xing/LAME header but applies them only with decoder gapless mode on, which was disabled so the manual `iTunSMPB` trim could not run twice. At a gapless boundary that inserted roughly 50 ms of silence (issue #1373).
- Trim ownership is decided per stream from what the demuxer actually reported, not from the codec name alone: iTunes-encoded MP3s report no encoder gap but do carry the tag, so keying on the codec would have left them with no trim at all.
- Ownership is not symmetrical. LAME writes its delay/padding extension independently of Xing's optional `FRAMES` field, and without a frame count the demuxer never produces a `trim_end` — such a file keeps the manual end trim while the decoder keeps the front, so neither end is cut twice.
- A packet that is trimmed away entirely is no longer read as end-of-stream, and an empty decode buffer reports a whole frame rather than a zero-length or infinite span.

## Test plan

- `cargo test --workspace --all-targets`: `psysonic-audio` 349/349. The only failure is the known wall-clock flake `repos::track::tests::upsert_500_rows_completes_well_under_perf_budget` in `psysonic-library` — green in isolation, and this branch does not touch that crate.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- Each behavioural test was verified red against the unfixed state by mutation: trimmed frame count and waveform phase, no double trim when an `iTunSMPB` tag is present, manual trim retained when the demuxer reports no gap, the streaming path, the deliberate exclusion of open-ended sources, a fully trimmed first packet through the resampler, seek accuracy, the frame alignment of an empty span, the end trim when the container omits its frame count (23087 frames before, 22050 after), and an exact frame count surviving an encoder symphonia does not recognise.
- Two tests are guards rather than regression tests, and say so in place: a header-less MP3 never reaches the seek clamp with an estimated frame count, and the bytes constructor needs no such filter because it picks its probe gate from sniffed bytes before the caller's hint.
- Fixtures are synthetic sine waves; `fixtures/README.md` carries the reference counts and the regeneration recipe. Two tests derive further shapes from them in memory (a Xing header without `FRAMES`, an unrecognised encoder string) rather than adding binaries.

### Runtime benchmark

Applicability: **required custom playback evidence** — the route harness exercises neither decoding nor track seams.

| Measurement | Before | After |
|---|---|---|
| Single track, decoded frames | 24192 | 22050 (= the encoded signal) |
| Two adjacent tracks at the boundary | 179712 | 176400, deviation 0 from the undivided source |
| Resampled 22.05 kHz MP3 | 0 (silence) | 47965 |
| Seek to start, frames remaining | 22003 | 22050 |
| LAME header without a Xing frame count | 23087 | 22050 |

Decode throughput, same probe on both sides, base is `origin/main` at `c2c89699` in a separate worktree with the fixtures copied in, one machine, release profile, runs taken sequentially:

| | Base (24192 frames) | Head (22050 frames) |
|---|---|---|
| Median of 41 runs | 0.222 ms | 0.151 ms |
| Minimum | 0.191 ms | 0.146 ms |

The head decodes 2142 frames fewer, which is exactly the encoder gap `fixtures/README.md` documents, and pays no measurable cost per frame for the trim. Codec and rate: MP3 at 44.1 kHz stereo and 22.05 kHz mono (MPEG-2 Layer III), gapless metadata present in the Xing/LAME header.

### Follow-ups left open

- The streaming constructor has no manual trim at all — it never sees the raw bytes and so cannot read an `iTunSMPB` tag. A locally cached MP3 whose LAME extension reports a gap while Xing omits `FRAMES` therefore keeps its end padding on that path. Same limitation the changelog already states for `iTunSMPB`-only files, rather than a new one.
- Random-access streaming sources now report a container duration where the path previously reported none, which arms the seek-past-end clamp for every codec rather than only the trimmed ones. That matches what the bytes constructor has always done, and replaces a rejected seek with a clamped one.
- The bytes constructor still accepts end-of-media with nothing decoded as an empty source. Making it an error was built in an earlier round and withdrawn: it turned track skips and preview hover-aways into playback errors, because that path has no generation signal to tell cancellation from truncation.
- The end trim places an `iTunSMPB` total against the delay the decoder removed, which those two tags need not agree on. Only the remainder is skipped so the total lands on the same sample — not covered by a test, because both variants deliver the same frame count and only the phase moves, which a sine fixture cannot show.

The reporter's own files have not been played against this build, so the issue stays open for their confirmation.


cucadmuh (dismissed):
The core MP3 encoder-delay diagnosis is sound, and the in-memory fixture coverage is strong. However, I found a reproducible production-path regression: a random-access streaming MPEG-2 MP3 produces 0 frames when resampling is enabled because `new_streaming` accepts a fully trimmed empty first packet. There are also unresolved trim-ownership gaps for finite non-seekable tracks and iTunSMPB-only local/ranged files, plus a seek-retry packet mismatch.

Local verification: focused decoder tests 14/14 passed; clippy passed; full workspace run had 772 passed and two unrelated `psysonic-library` wall-clock budget failures. GitHub CI is green, but the PR currently conflicts with `main` in `CHANGELOG.md`.

Please address the blocking streaming/resampling failure and the inline findings before merge.

PR #1392, opened 2026-08-05
Fix startup window controls and lifecycle races

### LINKED ISSUE #1349: [Bug]: Close and minimize buttons can ignore clicks during startup on Linux
### What's wrong?

On Linux, the native close and minimize buttons sometimes ignore the first click while Psysonic is opening.

The problem is most noticeable when clicking a window control immediately after the splash or main window appears. Waiting until startup finishes makes the controls work normally.

### Steps to reproduce

1. Start Psysonic on GNOME Wayland with the native title bar enabled.
2. As soon as the window appears, click minimize or close.
3. Repeat several cold launches.
4. Sometimes the first click is ignored and the window remains open.

The close failure is especially visible when “Minimize to Tray” is enabled.

### What should happen?

The native close and minimize buttons should respond consistently from the moment they become visible.

### What actually happens?

Clicks can be lost during the startup window. A later click usually works.

### Psysonic version

1.50.0. The same startup ordering is still present on current `main` / 1.51.0-dev.

### Operating system

Linux

### Install source

Built from source

Not relevant

### Logs

No application error is logged when the click is ignored.

### Anything else

Test environment:

- Bluefin / Fedora 44
- GNOME 50.3
- Wayland
- WebKitGTK 2.52.5
- Native GNOME title bar
- “Minimize to Tray” enabled

Source inspection found two startup ordering problems consistent with the behavior:

1. `src-tauri/src/lib.rs` always calls `prevent_close()` and emits
   `window:close-requested`. The frontend listener in
   `src/app/tauriBridge/useMediaAndWindowBridge.ts` is registered only after
   numerous sequential awaited media and shortcut listeners. Meanwhile,
   `public/startup-splash-reveal.js` and `src/app/startupSplash.ts` can show the
   window before React and the event bridge are ready. A close event received
   in that interval is prevented by Rust but has no frontend consumer.

2. On Linux, `src-tauri/src/lib.rs` initially disables window decorations for
   every main window. After React mounts, `usePlatformShellSetup.ts` enables
   them again when the native title bar is selected. Re-enabling GTK
   decorations restacks the window and temporarily changes focus, creating a
   second interval where native title-bar actions can be lost.

Suggested direction:

- Register the window lifecycle listeners before revealing the native window,
  using an explicit frontend-ready handshake.
- Resolve the initial Linux decoration mode before the first `show()` instead
  of switching `false → true` after the window is visible.
- Avoid pausing rendering on a close request until a consumer is guaranteed to
  handle it.

cucadmuh:
## Summary
- keep native window controls available through startup until the custom Linux title bar is committed
- replace the hidden-window animation-frame reveal path with generation-fenced timer/native lifecycle coordination
- queue close and force-quit requests safely across frontend loading, listener retries, native fallback, tray, mini-player, and second-instance restores
- prevent stale visibility or decoration operations from overriding newer user intent

Closes #1349.

## User impact
The app no longer strands users in a startup window without usable controls, and close, tray exit, restore, or mini-player actions are not lost during startup or a webview reload.

## Tauri contract
- added lifecycle commands for generation, listener registration, fallback policy, transition-fenced hide, and startup visibility
- changed `window:close-requested` to carry a transition token
- updated Rust handlers, generated bindings, frontend facades, listeners, and tests together
- no new high-frequency event stream was added; lifecycle events are emitted only for user/window actions

## Verification
- `npm run lint`
- `npm run dep:check`
- `npm test`
- `npx tsc --noEmit`
- `npx vitest run --coverage`
- `cargo test --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- --test-threads=1`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings`

All passed after merging current `origin/main`.

## Runtime benchmark
- Applicability: required; startup, application shell, shared listeners, and window lifecycle changed
- Baseline: `c1e4b01a9d69e0c6cc2950b9560d119b19d736ad` / report `2026-08-11T19-55-03-220Z`
- Final: `e0df3abd82a49ee4462d493573e31cc1a08df1f6` / report `2026-08-11T19-58-16-692Z`
- Confirmation rerun: `e0df3abd82a49ee4462d493573e31cc1a08df1f6` / report `2026-08-11T19-59-11-721Z`
- Command: `psysonic benchmark run --scenario core-pages --runs 2 --profile realistic --json`
- Environment: matching `1275x736` viewport at DPR 2, user agent, four selected servers, and library scope fingerprint
- Result: all routes reached, with no errors, timeouts, or skips; no long tasks were recorded
- The first final run's Albums and Favourites outliers did not reproduce. The confirmation run was at or faster than baseline on Albums, Artists, Tracks, and Favourites. Home warm total was `2238 ms -> 2286 ms` (+48 ms, 2%); its larger median movement came from a non-semantic first-sample DOM-quiet outlier that was absent from the first final run.

## Manual review notes
- `core.rs` exceeds the soft file-size trigger because the lifecycle state machine, commands, and invariant-focused tests are colocated; splitting the private state transitions would make race review harder.
- `cargo fmt` is unavailable in the current flake shell; Clippy and `git diff --check` pass.


Psychotoxical:
Independent second review pass on `8745abf4`.

Ran the full rule `09` gate set locally on Linux — lint, dep:check, tests, release notes, tsc, coverage, hot-path coverage, `cargo test --workspace --all-targets`, and clippy with `-D warnings`. All green; the Rust result lines were read per suite rather than trusting the exit code alone.

Also verified the two behaviours no test covers, on KDE Plasma / Wayland (non-tiling, custom title bar on), against a development binary confirmed to carry this branch: the splash window now keeps native controls and hands over to the custom title bar in one visible transition, and a close click during loading is queued and executed once the frontend reports ready — the process exited, confirmed from the process list.

No correctness findings. Two non-blocking notes:

1. `resumeRendering()` in `src/lib/api/platformShell.ts` has no call site — worth wiring up or dropping.
2. The reveal path lost its second route: it is now a three-step first-party chain, and if the generation lookup fails nothing is shown at all. That is deliberate and pinned by a test, and the fencing rationale holds — but with no native `show()` fallback, recovery depends on the tray entry or a second launch. Worth carrying as a stated residual risk rather than an assumption.

Approving from my side; merge stays with you.

Psychotoxical (dismissed):

Issue #1071, opened 2026-06-12
[Bug]: Prevents the computer from sleeping on Windows


Thraka:
### What's wrong?

_I don't know the exact terminology here for the power system_

The app seems to register and lock the audio stream/device on the system, which prevents the computer from sleeping. I noticed that my computer wasn't asleep after many hours of sitting there doing nothing. Psysonic was opened but music was paused. I checked `powercfg /requests` and noticed there was an audio stream in use. I closed the app and checked `powercfg` again and the audio stream preventing the computer from sleeping was gone. I then retested by opening Psysonic again and running `powercfg`. It seems to be blocking the system.

### Steps to reproduce

1. Close the app.
2. Run (admin) `powercfg.exe /requests` and note that there is nothing blocking (unless you have something else :grin:)
3. Open the app.
4. Don't play anything.
5. Run (admin) `powercfg.exe /requests` and note that the audio system is busy.

```
C:\Users\thraka>powercfg /requests
DISPLAY:
None.

SYSTEM:
[DRIVER] USB Audio Device (USB\VID_19F7&PID_0035&MI_00\6&176da87b&0&0000)
An audio stream is currently in use.

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.
```

### What should happen?

The app shouldn't be locking a device unless it's actually playing something.

### What actually happens?

The app locks the audio device even if nothing is playing.

### Psysonic version

1.47.0

### Operating system

Windows

### Install source

.msi installer (Windows)

### Subsonic-compatible server

Navidrome

### Server version

0.62.0

### Logs

```shell

```

cucadmuh:
Thanks for reporting this - that was really helpful.

What’s going on: Psysonic keeps the audio output open while the app is running so playback can start quickly, resume smoothly after a pause, and recover reliably when devices change or the system wakes from sleep. That’s why the audio device can still look “in use” even when nothing is playing.

That said, this isn’t the behavior we want. If music is paused (or you’ve only opened the app), the player shouldn’t stop your PC from sleeping. We’re treating this as a bug and will fix it by releasing the audio output after nothing has been playing for a while. We’ll use a one-minute idle delay before releasing it - long enough to avoid bouncing (repeatedly opening and closing the stream during short pauses), but short enough that the system can sleep when you’re not actually listening.

We’ll update this issue when a fix is available. Thanks again for taking the time to dig into it.

[...]

PR #1398, opened 2026-08-07
Fix Linux playback speed and queue advancement

cucadmuh:
## Summary
- verify the sample rate ALSA actually negotiates before configuring the Rodio mixer, preventing half-speed or double-speed playback
- keep requested and negotiated stream rates separate so Hi-Res and device recovery do not repeatedly reopen coerced streams
- hand off generation-qualified source completion state during Hi-Res and gapless source replacement so playback advances reliably
- preserve verified fallback coverage across Linux output devices and supported configurations

## Verification
- `nix develop -c bash -lc 'cargo test --workspace --all-targets -- --test-threads=1'`
- `nix develop -c bash -lc 'cargo clippy --workspace --all-targets -- -D warnings'`
- all 341 `psysonic-audio` tests pass
- default-parallel workspace runs expose two unrelated library timing-budget tests on this machine; both pass in the single-threaded workspace run and isolated runs
- independent static review found no remaining high/medium issues

## Manual verification
- play adjacent 44.1/48/96 kHz tracks with Native Hi-Res enabled on Linux ALSA/PipeWire
- confirm pitch and playback speed remain correct
- confirm gapless playback and ordinary queue advancement continue to the next track
- switch output devices and resume after an idle stream release

## Tauri boundary
- no invoke commands, events, or payload shapes changed

## Runtime benchmark
- not applicable: the route benchmark harness does not exercise native audio output; this change is limited to ALSA stream negotiation and playback completion correctness

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,6 +385,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Homepage URLs supplied by Navidrome no longer disappear from station cards or the edit form after saving or reloading the station.
 
+### Linux playback — keep speed, pitch and queue advancement correct
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#1398](https://github.com/Psychotoxical/psysonic/pull/1398)**
+
+* ALSA and PipeWire devices that select a different sample rate no longer play tracks at half or double speed with shifted pitch. Native Hi-Res remains available, and fallback output configurations verify the rate the device actually accepted.
+* Hi-Res and gapless source replacement no longer loses the track-completion signal, so playback advances to the next queued track reliably.
+
 ## [1.50.0]
 
 ## Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -389,7 +389,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1398](https://github.com/Psychotoxical/psysonic/pull/1398)**
 
-* ALSA and PipeWire devices that select a different sample rate no longer play tracks at half or double speed with shifted pitch. Native Hi-Res remains available, and fallback output configurations verify the rate the device actually accepted.
+* ALSA and PipeWire devices that select a different sample rate no longer play tracks at half or double speed with shifted pitch. Native Hi-Res remains available, and the selected output configuration verifies the rate the device actually accepted.
 * Hi-Res and gapless source replacement no longer loses the track-completion signal, so playback advances to the next queued track reliably.
 
 ### Servers — prevent duplicate account profiles
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -4208,6 +4208,7 @@ dependencies = [
 name = "psysonic-audio"
 version = "1.51.0-dev"
 dependencies = [
+ "alsa",
  "biquad",
  "dasp_sample",
  "futures-util",
--- a/src-tauri/crates/psysonic-audio/Cargo.toml
+++ b/src-tauri/crates/psysonic-audio/Cargo.toml
@@ -34,6 +34,7 @@ pitch_shift = "2.1"
 libc = "0.2"
 
 [target.'cfg(target_os = "linux")'.dependencies]
+alsa = "0.11"
 zbus = { version = "5.16", default-features = false, features = ["blocking-api", "async-io"] }
 
 [target.'cfg(windows)'.dependencies]
--- a/src-tauri/crates/psysonic-audio/src/alsa_rate.rs
+++ b/src-tauri/crates/psysonic-audio/src/alsa_rate.rs
@@ -1,0 +1,103 @@
+//! Linux ALSA sample-rate negotiation guard for CPAL/Rodio output streams.
+
+use alsa::pcm::{Access, Format, HwParams, PCM};
+use alsa::{Direction, ValueOr};
+use rodio::cpal::traits::DeviceTrait;
+use rodio::cpal::{Device, SampleFormat, SupportedStreamConfig};
+
+fn alsa_format_candidates(format: SampleFormat) -> Option<(Format, Option<Format>)> {
+    Some(match format {
+        SampleFormat::I8 => (Format::S8, None),
+        SampleFormat::U8 => (Format::U8, None),
+        #[cfg(target_endian = "little")]
+        SampleFormat::I16 => (Format::S16LE, Some(Format::S16BE)),
+        #[cfg(target_endian = "big")]
+        SampleFormat::I16 => (Format::S16BE, Some(Format::S16LE)),
+        #[cfg(target_endian = "little")]
+        SampleFormat::I24 => (Format::S24LE, Some(Format::S24BE)),
+        #[cfg(target_endian = "big")]
+        SampleFormat::I24 => (Format::S24BE, Some(Format::S24LE)),
+        #[cfg(target_endian = "little")]
+        SampleFormat::I32 => (Format::S32LE, Some(Format::S32BE)),
+        #[cfg(target_endian = "big")]
+        SampleFormat::I32 => (Format::S32BE, Some(Format::S32LE)),
+        #[cfg(target_endian = "little")]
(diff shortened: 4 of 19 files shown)


Psychotoxical:
Reviewed at `6133d63b`. I verified the premise against the pinned dependency rather than taking it on trust: cpal 0.17.3 `init_hw_params` (`src/host/alsa/mod.rs:1304`) does call `set_rate(..., Nearest)` and never reads the rate back, so the coercion this PR closes is real. `driver()` resolving to the PCM id checks out too, and `alsa_format_candidates` mirrors cpal's own format table including endianness gating and parameter order.

The completion rework looks right to me. Moving from the `AtomicBool` captured once in `audio_play` to a generation-qualified slot is what makes Hi-Res realignment survivable, and holding the slot lock across the generation check in `install_current_source_done` closes the overwrite race cleanly.

Two things I would like your read on before this merges.

**1. The Linux fallback now ends in `panic!` (`engine.rs:409`).** Every path is gated on `negotiated_output_rate`, which returns `None` at any of seven `.ok()?` points, and rodio's `open_default_sink()` last resort is gone. That panic runs on the stream thread, which has a single caller — when it unwinds, the reply sender drops, the caller sees only `audio stream open timed out`, and every subsequent `Open`/`Release` send fails for the process lifetime. So the failure mode is permanent silence plus a misleading log line, not a degraded stream. I understand the trade-off (no sound beats wrong pitch), but it is narrower than "preserve verified fallback coverage" suggests. Worth noting `engine.stream_handle` still holds an `Arc` clone of the old sink across a reopen — the stream thread drops only its own reference and the engine slot is replaced after the open succeeds — so on an exclusive `hw:` device the probe may hit a device cpal has not released yet.

**2. The probe opens run inside a device enumeration loop** (`open_verified_device_fallback`, and the sweep over `host.output_devices()` at `engine.rs:393`). cpal 0.17.3 dropped that pattern in its own backend for a documented reason — the comment above `supports_input` in `src/host/alsa/mod.rs`:

> Override trait defaults to avoid opening devices during enumeration.
>
> ALSA does not guarantee transactional cleanup on failed `snd_pcm_open()`. Opening plugins like alsaequal that fail with EPERM can leak FDs, poisoning the ALSA backend for the process lifetime (subsequent device opens fail with EBUSY until process exit).

That is the same condition that would drive point 1 to the panic, so the two compound rather than stack independently.

Minor: `maps_formats_used_by_cpal_alsa_output` only asserts `is_some()`, so a transposed pair in the mirrored table would still pass — and drift against cpal's table is the realistic defect for a hand-copied mapping.

Gates on my side: `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test --workspace --all-targets` had one failure, the known `upsert_500_rows_completes_well_under_perf_budget` wall-clock flake in `psysonic-library`, green on an isolated rerun and in a crate this PR does not touch.

**Caveat on all of the above:** I ran this on Windows, where `alsa_rate.rs` and the new `engine.rs` branches do not compile at all — 324 `psysonic-audio` tests here against your 341. Both points are read from the code, not measured. If they do not match what you see on real ALSA hardware, your observation wins.

Also note the branch conflicts with `main` on `CHANGELOG.md` only; #1399, #1400, #1401 and #1402 touch no audio file.

Psychotoxical:
Re-reviewed at `4f2bfa54`. All four points from the previous round are closed: the Linux path returns `Result` end to end and the stream thread forwards the error instead of unwinding, the enumeration-wide probe sweep is replaced by one config on one device, the engine slot drops its `Arc` before the reopen sequence starts, and the format test now compares the concrete pairs including endianness.

I traced the new synchronization layer across every call site. The ordering is consistent — `stream_open_lock`, then the attachment wait, then `playback_commit_lock` — and `connect_new_player` releases the open lock before its caller takes the commit lock, so the guard-then-commit sequence in the six callers does not close a cycle. The reentrancy case I went looking for is avoided cleanly: `reopen_output_stream` confines the lock to the `spawn_blocking` closure and calls `try_resume_after_device_change` only after it returns.

One thing worth your call, and one observation.

**`release_output_stream_on_stop` waits unbounded** (`stream_idle.rs:98`) while holding the open lock, and unlike `release_output_stream` it has no `output_stream_is_needed` precheck (`stream_idle.rs:79`). The attachment it waits on is held in `audio_play` from `connect_new_player` (`commands.rs:491`) through `drop(stream_attach)` (`commands.rs:591`), and that span contains `try_seek` (`commands.rs:551`) — an HTTP range request on network sources — plus a prefill sleep of up to 800 ms. `audio_stop` is the only caller and it is a command, so a stop can block for the length of that request. Not a deadlock: `audio_play` proceeds and the wait resolves. A bounded wait would cap it, with a timeout treated as "attachment still pending" so the release is skipped rather than racing it.

**Observation, not a defect:** `power_resume.rs` takes the commit lock (line 42) before the attachment wait it reaches through `request_stream_release_locked` (line 46), inverting the order every other path uses. It is safe today because `output_stream_is_needed` returns true while an attachment is pending (`stream_idle.rs:18`), and the held open lock stops a new attachment from starting between that check and the wait — so the wait is only ever reached at zero. Noting it because the safety there rests on that predicate rather than on the lock ordering.

Same caveat as the last round: this is read, not measured. `alsa_rate.rs` and the Linux branches in `engine.rs` do not compile on my machine. CI is green on both platforms.

cucadmuh:
Addressed the latest review at `e169dbda`:

- `audio_stop` now bounds the attachment wait to 50 ms. On timeout it still takes `playback_commit_lock`, verifies the stop generation, and stops already-registered preview/crossfade sinks; it skips only the device release, leaving the pending attachment race-safe for the idle watcher.
- The stop path rechecks the attachment count after acquiring the commit lock, so an attachment that finishes during the handoff still permits immediate release.
- Stream release was split into wait and post-wait helpers to avoid re-entering an unbounded wait.
- System resume now follows `stream_open_lock -> attachment wait -> playback_commit_lock`, with the needed-state rechecked under the commit lock.
- Added Condvar tests for notification success and bounded timeout.

Verification on the final commit:
- `cargo test --workspace --all-targets -- --test-threads=1`
- focused workspace `stream_idle` tests: 11 passed
- independent concurrency re-review: no blocking or medium findings

Runtime benchmark: not applicable - this changes native audio-device synchronization, outside the route benchmark harness.

Psychotoxical:
Re-reviewed at `e169dbda`. The bounded-wait point is closed, and closed conservatively: on timeout the stop still takes `playback_commit_lock` and verifies the stop generation, giving up only the device release.

I checked the ordering that makes the unconditional teardown safe. `audio_play` bumps the generation under the commit lock before any work (`commands.rs:156-159`), so an attachment pending at stop time always carries a generation below `stop_generation` and aborts at `commands.rs:575`; a play starting after the stop bumps past it and the release returns before the teardown. The recheck at `stream_idle.rs:114` is sound because the open lock is held throughout, so the pending count can only fall. The helper split is complete — the waiting variant has one caller left and it is guarded by `output_stream_is_needed`.

No blocking or medium findings. Three notes, all optional:

- The skip is logged with `app_deprintln!` while the success line beside it is `app_eprintln!`. That branch is the one users report as "device still in use after stop", and it is the only trace of it.
- `bounded_attachment_wait_succeeds_after_notification` gives a 5 ms handoff a 100 ms budget. Raising it removes a future CI flake at no cost — the wait returns on the notification, not the timeout.
- `power_resume.rs:48` returns when the recheck finds the stream needed, rather than continuing to the reopen below it. Matches `main`, so not a regression.

Worth knowing for issue #1071 reports: in the timeout case the device stays open until the idle watcher gets to it. Still the right trade against an unbounded wait.

Caveat as before: read, not measured — the Linux branches do not compile on my machine. All six checks green on the head commit, including `cargo test --workspace (windows)`.

cucadmuh:
Accepted the two actionable optional notes in `7a2141e5`:

- the attachment-timeout skip now uses `app_eprintln!`, so issue #1071 reports have a normal runtime trace when Stop leaves the device open for the idle watcher;
- the notification test budget is now 1 second, while still returning immediately on the Condvar notification, removing avoidable CI timing sensitivity.

Left the `power_resume` early return unchanged as noted: it matches existing behavior and is not a regression.

Verification:
- focused workspace `stream_idle` tests: 11 passed
- `cargo test --workspace --all-targets -- --test-threads=1`

Runtime benchmark: not applicable - diagnostics and test-stability only.

Issue #1408, opened 2026-08-13
[Bug]: 5.1 FLAC playback loses accompaniment on stereo output


tossmilestone:
### What's wrong?

Psysonic 1.50.0 on macOS cannot correctly play some multichannel FLAC files when the output device is stereo.

For a 5.1 FLAC file, vocals/dialog are audible, but most of the accompaniment/instrumental content is missing.

The same file plays correctly with `ffplay`, so the source FLAC appears to be valid and FFmpeg handles the 5.1 → stereo conversion correctly.

### Steps to reproduce

The problematic file reports:

```text
Stream #0:0: Audio: flac, 44100 Hz, 5.1(side), s16
Stream #0:1: Video: mjpeg (Baseline), yuvj444p(pc, bt470bg/unknown/unknown), 500x500 [SAR 1:1 DAR 1:1], 90k tbr, 90k tbn (attached pic)

Metadata:
  comment : Cover (front)
  title   : Front Cover
```

Playing the original file with:

```bash
ffplay input.flac
```

works correctly — both vocals and accompaniment can be heard.

Playing the same track in Psysonic results in:

* vocals/dialog are audible
* accompaniment/instrumental content is largely missing

### What should happen?

A 5.1 FLAC should be correctly downmixed to the available stereo output device, preserving the musical content from all relevant channels.

### What actually happens?

The multichannel track appears to be played without correct 5.1 → stereo channel mapping/downmixing.

### Psysonic version

1.50.0

### Operating system

macOS

### Server version

Navidrome 0.58.0

### Logs

```shell

```

cucadmuh:
https://github.com/Psychotoxical/psysonic/releases/tag/app-v1.51.0-rc.1

PR #1413, opened 2026-08-14
fix(audio): mix multichannel tracks down to stereo instead of dropping channels

Psychotoxical:
## Summary

- A multichannel track played through a stereo device lost most of its content. rodio converts every source to the mixer's channel count, and where a source has more channels than the device it keeps the first ones and discards the rest — so of a 5.1 track's six channels only front-left and front-right were ever heard, with centre, LFE and both surrounds silenced (issue #1408). Measured on the mixer's own converter: fed channels 1 through 6, it returns 1 and 2.
- Sources are now folded to stereo before the mixer sees them, using the gains the spectrum tap already describes these layouts with — one model for what is shown and what is heard. The fold sits directly after resampling, so EQ, fades, the visualiser and the sample counter all work on the channels that will be played, and the reported channel count, which the playback position derives from, matches the output.
- A device that takes the source's channels is left untouched: folding unconditionally would fix stereo playback by taking surround away from the people who have it.
- Attenuation follows the power sum (about -3.5 dB for 5.1) rather than the amplitude worst case, which would have put 5.1 tracks roughly 9 dB below the stereo tracks around them in a queue. Correlated material is clamped.

## Test plan

- `cargo test --workspace --all-targets --no-fail-fast`: `psysonic-audio` 365/365. The only failure is the known wall-clock flake `repos::track::tests::upsert_500_rows_completes_well_under_perf_budget` in `psysonic-library` — green in isolation, and this branch does not touch that crate.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- The defect and the fix are measured against a real 5.1 FLAC, not a stand-in: `fixtures/five_one_sine.flac` carries a different tone per channel, so the stereo output can be asked which channels reached it. All six are present, each surround stays on its own side, and neither leaks into the other. Before the fix only the front pair survived.
- Channel indices come from symphonia's own FLAC channel map. 6.1 is the one layout that does not continue the left/right pairing (FL FR FC LFE **RC** SL SR) and is covered separately from the layouts that do.
- Guards for the `rodio::Source` contract: an exhausted span stays exhausted (rodio defines `is_exhausted()` as exactly `current_span_len() == Some(0)`, and its queue looks ahead with it at a gapless boundary), spans stay a whole number of frames, and a seek keeps the sample still owed — dropping it shifts interleave parity by one and swaps left and right for the rest of the track.

### Runtime benchmark

Applicability: **required custom playback evidence** — the route harness exercises neither decoding nor channel conversion.

| Measurement | Before | After |
|---|---|---|
| 5.1 source, channels reaching a stereo device | 2 of 6 (front pair only) | 6 of 6 |
| Surround placement | n/a — discarded | left surround left, right surround right, no cross-leak |
| 5.1 source on a 5.1 device | 6 | 6 (unchanged) |
| Attenuation applied to a 5.1 fold | none (nothing was summed) | -3.5 dB, clamped against full scale |

Per-sample cost is one multiply-add per channel, replacing a discard; no allocation, no extra buffering beyond the one sample of a folded frame.

## Not covered

- **Nobody has heard it.** Neither maintainer has 5.1 material or a surround device; every measurement here is synthetic. The reporter is the one who can confirm the result on their own file and setup.
- **Internet radio is unchanged.** That path assembles its source chain directly rather than through `build_source`, so a multichannel radio stream still goes through rodio's conversion. Stations are effectively always stereo, so this is a gap in coverage rather than a live defect.
- A device that takes more than two channels but fewer than the source (a 4.0 output fed 5.1) keeps rodio's behaviour — that needs its own layout mapping, not this one.


cucadmuh (approved):
lgtm

PR #1425, opened 2026-08-16
Add configurable force scrobbling

JayDawgThaGOAT:
### Changes

Settings → Integrations → Music Network now has a 25–90% scrobble threshold (default 50%). That percentage also gates the media-server scrobble when no Music Network account is connected.

Force scrobble is a player-bar action (also in fullscreen and on the elapsed-time status card). It submits once per play to the owning server and every enabled Music Network destination. Radio, an already-scrobbled play, and offline browse are refused.

New locale keys use the English strings. Changelog/credits are included; I will add the PR number after this opens. If 1.51 is already frozen, I can move those to 1.52.0.

### Verify
- Threshold 25% scrobbles once at 25%; 50% does not fire at 25%
- Force before the threshold submits once, then shows the checked badge
- Radio / offline / already-scrobbled stay blocked
- Hide the action in the Player Bar customizer

Tauri boundary: unchanged.


cucadmuh:
Thanks for the PR. We'll review it as part of the 1.52 release cycle.

PR #1430, opened 2026-08-17
refactor(rust): split oversized modules

cucadmuh:
## Summary

- split oversized Rust modules into focused submodules across analysis, audio, integrations, library, sync, cache, CLI, startup, and app-host code
- keep existing public APIs, Tauri command/event contracts, SQL, migrations, and runtime behaviour unchanged
- move co-located tests with their implementation while preserving the same production paths
- reduce hand-written Rust files over 600 lines from 61 to 5; the remaining large files are cohesive or table-heavy exceptions documented in the task notes

## Verification

- `cargo test --workspace --all-targets -- --test-threads=1` - passed
- `cargo clippy --workspace --all-targets -- -D warnings` - passed
- `cargo check --workspace --all-targets` - passed
- `cargo check -p psysonic --release --all-targets` - passed
- `RUST_TEST_THREADS=1 cargo llvm-cov --workspace --json --output-path target/llvm-cov/cov.json` - passed
- `bash scripts/check-hot-path-coverage.sh` - passed (20/20 files above 70%, none missing; Discord presence helpers 100%)
- `committed_bindings_are_fresh` - passed
- `typeable_commands_are_all_collected` - passed
- `every_command_is_registered_in_the_handler` - passed
- `git diff --check` - passed

The default parallel workspace test run exceeds the timing budget in two 500-row ingest performance tests. The same two failures reproduce on `origin/main` at approximately 1.23-1.27 s under full workspace contention. Both tests pass individually on this branch at approximately 70-80 ms, and the complete serial workspace run passes. This is an existing local test-contention issue rather than a branch regression.

## Runtime benchmark

- Applicability: required because the refactor touches shared native runtime, browse, analysis, audio, cache, and startup modules
- Baseline: `cde28f54` / report `2026-08-17T09-53-15-644Z`
- Final benchmarked runtime source: `b7d8ba89` / report `2026-08-17T09-56-37-034Z` (`f06b8a82` updates CI coverage target paths; `8395e4b0` only moves unchanged Discord pure helpers into a gated child module)
- Command: `psysonic benchmark run --scenario all-pages --runs 2 --profile realistic --json`
- Environment: Linux debug, viewport `657x1165`, DPR 2, four selected servers, identical stable library-scope fingerprint
- Coverage: all static routes plus available data-backed detail routes, filters, advanced search, and pagination
- Result: no errors, timeouts, wrong routes, failed interactions, or long tasks

Semantic route readiness, baseline -> final:

| Route | Baseline | Final |
| --- | ---: | ---: |
| Albums | 262 ms | 215 ms |
| Favourites | 202 ms | 155 ms |
| Tracks | 705 ms | 652 ms |
| Artists | 473 ms | 468 ms |

Home readiness was noisy in the final pair (`675 -> 1137 ms`), but a second head run in the identical environment measured `661 ms`; the increase did not reproduce. Filter, search, and pagination semantic timings remained comparable. Generic totals on routes without semantic readiness varied with cover/network DOM-quiet work and did not show a reproducible branch-specific regression.

Expected incomplete coverage:

- `/label/:name` skipped because the configured library had no reachable owned album with a record label
- lossless-album and advanced-search pagination skipped because no additional page was available

Local benchmark reports were inspected but are not committed because they contain environment and library-derived data.

## Tauri boundary

- no command names, event names, argument shapes, payloads, or error contracts were added, removed, or changed
- bindings freshness and command-registration tests passed

## Release notes

- Added a compact `CHANGELOG.md` entry under `1.52.0` → `Changed` for the large internal module split and its verification.
- No `WHATS_NEW.md` or contributor-credit entry: this remains an internal code-placement refactor with no intended user-visible feature.


Psychotoxical:
Reviewed at `b7d8ba89` for code, `f06b8a82` for the CI follow-up. **Approve** — no blocking findings, one question at the end.

This is a move refactor, so the review targets the claim directly: same code, different files. Every `.rs` file under `src-tauri` was compared between base and head as sorted multisets of the things that would change if behaviour moved — not a sample, the whole tree. Each difference was then traced down to the individual function.

| Invariant | Base | Head | |
|---|---:|---:|---|
| Type definitions | 601 | 601 | identical |
| `#[tauri::command]` | 291 | 291 | identical |
| Test attributes | 1549 | 1549 | identical |
| `pub fn` | 1039 | 1039 | identical |
| Function definitions removed | — | 0 | nothing lost |

The three sets that differ all resolve cleanly:

* **`cfg` gates:** 39 `target_os = "linux"` gates disappear and 39 `cfg(test)` appear — consolidation, not loss. `dev_io.rs` carried 37 per-test gates; the tests now sit in `pipewire_tests` behind `cfg(all(test, target_os = "linux"))` and in `device_identity_tests` behind `cfg(test)` with 6 gates inside. 25 tests before, 25 after.
* **SQL:** same count, one line differs, same statement in a different call form.
* **Function names:** 27 added, 0 removed. 17 are the same name twice, where a forwarding wrapper keeps the old path while the body moves into the child module. The other 10 are the `_impl` cores behind unchanged commands and the new `startup/` entry points.

Two extra checks, because this moves file-deletion code: the containment guards (`resolve_within_root`, `planned_path_stays_within`, `is_path_on_mounted_volume`) are all still on the `sync/device/rename.rs` path, and visibility was widened the narrow way — `pub(super)` 40 → 582 and `pub(in …)` 0 → 55, with `pub(crate)` up only 888 → 934.

**Windows workspace run** (local, since the API was down during the review): `cargo check` and `cargo clippy --workspace --all-targets -- -D warnings` pass; `cargo test --workspace --no-fail-fast` runs all 8 crates with **1714 passing** — the same figure this workspace produces on `main`. The one failure is the known wall-clock flake in `upsert_500_rows_completes_well_under_perf_budget`, green in isolation at 0.06 s. Unrelated environment note for anyone repeating it: a worktree under a deep path fails in `opusic-sys` with `MSB3491 … exceeds the maximum path limit`; redirecting `CARGO_TARGET_DIR` to a short path fixes it.

---

**One question, on the coverage gate.** `compute/` and `decode/` are retargeted completely — every non-test file is listed. `discord.rs` leaves the gate entirely, and the mechanics behind that are sound: the split moved the coverable logic into `artwork.rs`, so what remains is 344 lines of which roughly 135 are the IPC function, and the per-file floor is no longer realistic.

Worth naming explicitly, though: four pure functions stay behind in `discord.rs` — `is_publishable_image_url`, `apply_template`, `compute_discord_text_fields`, `compute_discord_start_timestamp` — and 20 existing tests cover exactly those (9 in `security_tests.rs`, 11 in `text_tests.rs`, the URL check referenced 11 times). Those tests keep running; their subject is simply no longer gate-enforced, including a URL-publishability check.

Following the pattern already used here, extracting those four into their own file and gating that would keep the enforcement without demanding coverage of the transport. Your call — not a blocker either way.

Psychotoxical (dismissed):

cucadmuh:
Addressed in `8395e4b0`. I moved the four pure presence helpers (plus their private constant/result type) into `discord/presence.rs` and added that implementation file to the hot-path gate. The existing security/text tests are unchanged and still exercise the same functions through the parent module. Local verification: 37 focused Discord tests passed; `cargo test --workspace --all-targets -- --test-threads=1` passed; workspace Clippy passed; `presence.rs` is 100.0% covered and the gate passes 20/20 with no missing files. Thanks for calling out the enforcement gap.

Psychotoxical:
Checked `8395e4b0`: pure move. All four helpers now live in `discord/presence.rs`, function and test counts in `psysonic-integration` are unchanged at 233 and 70, and the file is in the gate. Question closed, approve stands.

PR #1449, opened 2026-08-23
fix: preserve OpenSubsonic artist refs during Navidrome N1 delta sync

devyeah1978:
## What changed

Navidrome's native `/api/song` payload used by N1 delta sync can omit richer OpenSubsonic fields such as `artists`, `albumArtists`, and `displayArtist`.

Previously, N1 delta updates used the normal remap-aware upsert path, which replaced the existing `raw_json` wholesale. This could erase structured artist references previously captured from richer OpenSubsonic responses.

The result is that a track such as:

- `artist = "FOVOS, Max Cardona"`
- `artists = [{FOVOS}, {Max Cardona}]`

could later lose `artists[]` after an N1 delta refresh. The frontend then falls back to the flat display string, and because commas are intentionally not treated as artist separators, the two artists appear as one combined artist.

This change:

- adds a sparse remap-aware upsert path
- uses it for Navidrome N1 delta ingest
- preserves existing JSON fields that are absent from the native delta payload via the existing `json_patch` behavior
- adds a regression test reproducing the multi-artist case

## User impact

Navidrome users with multi-artist tracks should no longer see structured artist credits collapse into a single combined artist after background/delta synchronization.

## How to verify

1. Start with a locally indexed track whose `raw_json` contains structured `artists[]`.
2. Run an N1 delta where `/api/song` returns the same track with only the flat `artist` string.
3. Confirm the updated row still contains the original structured `artists[]`.
4. Confirm the UI continues to show individually resolved artists.

The included regression test reproduces this with `FOVOS` and `Max Cardona`.

## Scope

Backend/Rust only.

- No frontend changes.
- No Tauri command/event contract changes.
- No schema or migration changes.
- No changes to server-side metadata.
- Existing sparse-upsert behavior is reused rather than introducing comma-based artist splitting.

## Testing

Added an N1 delta regression test covering preservation of `artists[]` and `albumArtists[]`.

Upstream CI should run:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Artists, albums and songs could each appear twice when the local index and the server response used different forms of the same server identity. Live Search now aligns that identity before combining the results, while still keeping genuinely separate matches from different servers.
 
+### Navidrome multi-artist credits survive background sync
+
+**By [@devyeah1978](https://github.com/devyeah1978), PR [#1449](https://github.com/Psysonic/psysonic/pull/1449)**
+
+* Navidrome delta updates no longer collapse structured track and album artist credits into one comma-joined artist. Current native credits replace stale names while richer OpenSubsonic artist references remain intact when the native response omits them.
+
 ### Linux shortcuts work immediately after returning to the app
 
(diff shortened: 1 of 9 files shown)


devyeah1978:
Quick CI note: both cargo test --workspace jobs and coverage passed. The Linux/Windows Clippy failures are on pre-existing psysonic-syncfs/src/sync/batch.rs code under Rust 1.98 (clippy::needless_late_init), not on the N1 sparse-upsert changes in this PR. Current main already contains the Clippy cleanup, so I’m updating this branch onto current main and will rerun CI against that baseline. No behavior change is being made to the N1 artist-preservation fix itself.

PR #1470, opened 2026-08-28
fix(library): restore Navidrome timestamps with UTC offsets

cucadmuh:
## Summary

- Parse positive and negative RFC 3339 UTC offsets correctly in Subsonic and Navidrome timestamps.
- Preserve creation timestamps when sparse initial-ingest or resync payloads omit them.
- Run a bounded, resumable background reconcile that restores only server timestamps demonstrably damaged by the old parser, without overwriting newer favourite/play state.

## User impact

Navidrome libraries whose timestamps contain negative UTC offsets can populate New Releases again. Safely identifiable existing values are repaired gradually after startup without blocking database open; ambiguous legacy values wait for a later sync rather than risk restoring intentionally cleared data.

## Verification

- `cargo clippy --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- -D warnings`
- `cargo test --manifest-path src-tauri/Cargo.toml --workspace --all-targets -- --test-threads=1`
- Timestamp, remap, and scheduler regression tests cover offset parsing, authoritative hot-column preservation, explicit clears, existing remap destinations, raw metadata non-resurrection, bounded resume, and drain/cancellation behavior across non-abortable blocking work.
- Real database A/B using a supplied 11,397-track snapshot:
  - Base: 11,397 missing creation/update timestamps; New Releases returned 0 albums.
  - Head: 0 missing creation/update timestamps; New Releases returned 30 albums.
  - The original synchronous implementation completed open and repair in 885 ms; the final implementation removes that O(n) work from database open.
- The exact parallel workspace test command hits the same two timing-only ingest assertions on base and head under local contention: base 1.16/1.20 s, head 1.18/1.22 s. Both pass in isolation in 0.07-0.08 s, and the complete serial workspace run passes.
- Follow-up isolated repair on an 11,719-track pre-fix copy:
  - `server_created_at` NULL: 11,719 → 0.
  - `server_updated_at` NULL: 11,719 → 2; both remaining payloads have no update field.
  - Starred rows: 0 → 0; played rows: 17 → 17, confirming the repair does not rebuild mutable user state from stale JSON.
- Marker completed at cursor 11,722.
- Database-open regression coverage confirms the repair marker is not created or drained by the open pipeline.
- Final serial workspace result: 121 root tests, 186 analysis tests, 398 audio tests, 52 core tests, 126 integration tests, 830 library tests, and 159 syncfs tests passed; 15 tests were ignored. Clippy passed with `-D warnings`.
- Two independent final static review passes reported no findings after the data-provenance and scheduler-lifecycle fixes.

## Runtime benchmark

- Applicability: required; this removes database-open work, adds low-priority scheduler work, and changes data used by the Home New Releases feed.
- Baseline: `6d88c4ac6aafc233b1a29cccdc85998647aa71cf` / report `2026-08-28T09-56-53-612Z`
- Original final code: `07489ba9` / report `2026-08-28T10-01-08-116Z`
- Original confirmation: report `2026-08-28T10-03-09-731Z`
- Data-safety commit: `92d4d8bd` / reports `2026-08-28T12-36-20-012Z`, `2026-08-28T12-36-58-570Z`, `2026-08-28T12-37-41-813Z`
- Background final: `8b01e51e` / reports `2026-08-28T14-13-06-983Z`, `2026-08-28T14-13-47-743Z`, `2026-08-28T14-18-50-017Z`, `2026-08-28T14-19-32-759Z`
- Final reviewed head: `6390802d` / reports `2026-08-28T17-44-50-936Z`, `2026-08-28T17-45-43-798Z`
- Command: `psysonic benchmark run --scenario core-pages --runs 2 --profile realistic --json`
- Environment: Linux debug, one selected server, stable scope fingerprint. Baseline/background comparison used 1283x694 at DPR 2; the repeated final-head smoke pair used 1269x694 at DPR 2.
- Affected route: Home `/`, including the New Releases feed.
- Result: no errors, timeouts, skipped routes, or wrong routes. The repeated final-head run reported Home 1,222 ms, Albums 1,055 ms, Artists 841 ms, Tracks 996 ms, and Favorites 1,632 ms median total; the first-run Tracks spike did not reproduce.
- Noise check: Home alternated between approximately 0.9-1.4 s route work and a 6-7 s post-ready DOM-quiet wait. The spike is not semantic readiness, does not reproduce on the other routes, and matches earlier scheduler/background-noise reports. The repair emits no frontend event and the benchmark profile's repair marker was already complete.
- Errors, timeouts, wrong routes, and affected interaction failures: none.

## Database and API notes

- No schema migration or compatibility-version change.
- The repair uses the existing `library_data_migration` marker/cursor mechanism, reads exactly 1,000 physical rows per transaction, and writes only rows whose server timestamp is demonstrably affected.
- Database open does not run the repair. The existing 30-second library scheduler waits through its first tick, runs due sync sessions first, and then processes at most one repair batch only while playback is idle, no foreground sync or migration is active, and bulk ingest is inactive. The activity barrier remains held inside blocking work even if its async wrapper is cancelled, and app exit closes scheduler admission.
- No Tauri command, event, or payload contract changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,7 +188,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1470](https://github.com/Psysonic/psysonic/pull/1470)**
 
-* Navidrome timestamps with negative UTC offsets no longer disappear from the local index, so New Releases, favourites and last played dates populate correctly. Existing libraries are repaired once at startup without requiring a full resync.
+* Navidrome timestamps with negative UTC offsets no longer disappear from the local index, so New Releases, favourites and last played dates populate correctly during sync. Existing library creation dates are repaired once at startup without requiring a full resync or overwriting newer favourite/play state.
 
 ## [1.51.0] - 2026-08-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -184,6 +184,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Nine topics that were never covered are now there: browsing several servers at once, the timeline queue view with Play from Here, capping the streaming bitrate per server address, the table view on the album pages, telling your own playlists from shared ones, the Composers page, the audio visualizer, the theme store and the background sources.
 * All of it in every language the app ships in.
 
+### Navidrome dates with UTC offsets populate New Releases again
+
+**By [@cucadmuh](https://github.com/cucadmuh), PR [#1470](https://github.com/Psysonic/psysonic/pull/1470)**
+
+* Navidrome timestamps with negative UTC offsets no longer disappear from the local index, so New Releases, favourites and last played dates populate correctly. Existing libraries are repaired once at startup without requiring a full resync.
+
 ## [1.51.0] - 2026-08-17
 
 ## Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,7 +188,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 **By [@cucadmuh](https://github.com/cucadmuh), PR [#1470](https://github.com/Psysonic/psysonic/pull/1470)**
 
-* Navidrome timestamps with negative UTC offsets no longer disappear from the local index, so New Releases, favourites and last played dates populate correctly during sync. Existing library creation dates are repaired once at startup without requiring a full resync or overwriting newer favourite/play state.
+* Navidrome timestamps with negative UTC offsets no longer disappear from the local index, so New Releases, favourites and last played dates populate correctly during sync. Existing creation dates with safely identifiable old values are repaired gradually in the background; ambiguous cleared legacy values wait for a later sync rather than risk restoring stale data.
 
 ## [1.51.0] - 2026-08-17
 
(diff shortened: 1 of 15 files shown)


cucadmuh:
## Final review: request changes

CI is green and the batching direction is sound, but the pushed head `8b01e51e` still has three correctness issues:

1. **Explicit sparse clear can restore a stale alias** (`track_timestamp_reconcile.rs:122-130`, `ingest.rs:334-347`). SQLite `json_patch` removes `createdAt` when the patch contains `createdAt: null`; the reconcile can then fall back to an older retained `created` alias and repopulate the cleared hot column.
2. **Sparse remap can overwrite a newer existing destination** (`remap.rs:77-119`). Old-row timestamps are supplied as `excluded.server_*` even when the destination ID already exists, so the conflict update can replace newer destination values.
3. **Background repair is outside the scheduler drain barrier** (`library_scheduler.rs:21-50`, `runtime.rs:697-762`). The batch does not hold `sync_activity_guard()`, allowing `cancel_and_drain_sync()` to report the scheduler drained while repair is queued or running.

Also update `CHANGELOG.md:191`: repair is now gradual background work after startup, not a one-time synchronous startup repair.

Required tests:

- Sparse upsert with `createdAt: null`, followed by reconcile, with a retained old `created` alias.
- Old remap source plus an already-existing destination carrying newer timestamps.
- An exclusive sync drain waiting for an in-flight repair batch, including a final eligibility re-check after acquiring the activity guard.

Detailed review artifact: `psysonic-workdocs/internal/collaboration/handoffs/2026-08-28-pr-1470-review.md`.

PR #1490, opened 2026-09-02
fix(library): read Navidrome's native strong keys and keep profile ids out of the analysis server key

### LINKED ISSUE #1434: [Bug]: 1.51.0 — canonical_* tables stay empty after migrations 19–26, every enrichment write fails with "FOREIGN KEY constraint failed"
### What's wrong?

After updating 1.50.0 → 1.51.0 the new canonical layer added by migrations 19–26 is never populated: `canonical_track`, `canonical_identity`, `track_canonical_link` and `canonical_enrichment_link` all stay at 0 rows. As a result every analysis enrichment write fails with `FOREIGN KEY constraint failed` (the enrichment link references `canonical_track(id)`, which has no rows), and enrichment-backed data (waveform / loudness) never lands.

Cluster keys *are* built — `library-cluster.db → track_cluster_key` has a row for all 27441 tracks — so the identity stage runs, but the `canonical.link_all_tracks` step apparently never materialises any `canonical_track` row.

### Steps to reproduce

1. Existing library on 1.50.0: Navidrome, 27441 tracks, `library.sqlite` at schema version 18.
2. Update to 1.51.0 and launch — migrations 19–26 apply on first start.
3. Let it idle for ~20 s (or play a track) so analysis enrichment runs.
4. Watch stdout / the log.

### What should happen?

The canonical layer gets populated (a `canonical_track` row per identity, `track_canonical_link` rows for the tracks), and enrichment links insert successfully.

### What actually happens?

Enrichment fails on every track:

```
[analysis][enrichment] failed track_id=4q20JrA8MsH3mveDnO1qVr server_id=<server>: FOREIGN KEY constraint failed
```

DB state right after the 1.51.0 first run (schema version 26, `pragma foreign_key_check` is clean):

| table | rows |
|---|---|
| `track` | 27441 |
| `artist` | 1289 |
| `album_browse_projection` | 6161 |
| `cluster.track_cluster_key` | 27441 |
| `canonical_track` | **0** |
| `canonical_identity` | **0** |
| `track_canonical_link` | **0** |
| `canonical_enrichment_link` | **0** |

`schema_migrations`: versions 1–18 predate the update, 19–26 all recorded with the same timestamp of the 1.51.0 first launch. `library_data_migration` shows `duration_sec_decimal_backfill_v1`, `scope_browse_album_projection_v1` and `scope_browse_composer_projection_v1` completing over 27445 rows — no canonical backfill entry is present, so it is not obvious to me whether the linking pass is supposed to run as a data migration or lazily.

### Psysonic version

1.51.0

### Operating system

Linux

### Install source

Other (please describe in "Anything else")

Navidrome

### Server version

0.62.0 (1b46b977)

### Logs

```shell
(psysonic:48106): libayatana-appindicator-WARNING **: libayatana-appindicator is deprecated. Please use libayatana-appindicator-glib in newly written code.
[2026-08-18 11:19:53.822 +0300] [psysonic] logind PrepareForSleep listener registered (post-sleep audio recovery)
[2026-08-18 11:19:56.856 +0300] [analysis][enrichment] failed track_id=4q20JrA8MsH3mveDnO1qVr server_id=<server>: FOREIGN KEY constraint failed
```

From the first run, which applied the migrations:

```shell
[library-db] SLOW write op=identity.ensure_cluster_keys_built lock_wait_ms=0 exec_ms=1092
[library-db] write op=genre_tags.ensure_schema lock_wait_ms=363 exec_ms=0
[library-db] write op=browse_projection.backfill lock_wait_ms=0 exec_ms=715
[library-db] write op=composer_projection.backfill lock_wait_ms=0 exec_ms=400
[library-db] write op=identity.ensure_cluster_keys_built lock_wait_ms=0 exec_ms=667
[analysis][enrichment] failed track_id=h31z3LDk6n40AnUKkQBodT server_id=<server>: FOREIGN KEY constraint failed
```

### Anything else

- Bluefin / Fedora 44, Wayland, host WebKitGTK 2.52.5.
- Install source: official `Psysonic_1.51.0_amd64.AppImage`, with the binary extracted from the AppImage and run against the host WebKitGTK — the bundled WebKitGTK still aborts here, see #1347. Same setup worked fine on 1.50.0.
- I have rolled back to the 1.50.0 binary on the already-migrated (version 26) database; happy to run any diagnostic build against a copy of this DB.
- Possibly related: #1431 also appeared in 1.51.0 and also touches the new library-scope layer. In my DB the browse projection is populated under `library_id = '1'`, while the single `sync_state` row has `library_scope = ''` — I don't know whether that mismatch is expected, but mentioning it in case the two share a root cause.

Psychotoxical:
## Summary

- The analysis boundary refuses only ids the app itself minted, tracked in `mintedServerProfileIds`; shape is not consulted. An address-derived key stays valid there even when its server profile is gone, so those tracks keep waveform and loudness.
- Analysis enrichment could write `track_fact` rows under a generated profile id the library does not know, causing repeated `FOREIGN KEY constraint failed` errors. Generic storage-key resolution keeps every address-derived and persisted key unchanged; the playback analysis boundary alone refuses an unresolved id that the app itself minted.
- The native Navidrome mapper now reads `mbzRecordingID` and `tags.isrc`, so natively ingested libraries populate the existing `mbid_recording` and `isrc` hot columns.
- One-time resumable backfill `native_strong_keys_backfill_v1` fills empty strong-key columns from `raw_json` and links every live keyed row lacking `track_canonical_link`, in 1,000-row batches from the idle scheduler. Populated hot columns are never rewritten.

Closes #1434

## Test plan

- [x] `npm run lint`, `npm run dep:check`, `npm run prebuild:release-notes`, `npx tsc --noEmit`, full Vitest, coverage, and the frontend hot-path gate
- [x] `cargo test --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`
- [x] Resolver and analysis tests cover profile UUIDs, unconfigured bare hostnames of any base36 shape, a minted id whose profile is no longer configured, and an address-derived key surviving profile removal; rehydrate tests cover seeding from configured servers, retention of removed ids, and the no-op case
- [x] Native mapper tests cover `mbzRecordingID`, `tags.isrc`, fallbacks, and blank values
- [x] Backfill tests cover fill/link behavior, hot-column precedence, tombstones, unkeyed rows, resumable batches, completion idempotence, and exclusion from database open
- [x] Reporter database snapshot (27,441 live native tracks) drained on a copy: strong-key columns 0 -> 13,356, `track_canonical_link` 0 -> 13,356, `canonical_track` 0 -> 12,689; 29 debug batches at 250-535 ms; a further pass was a no-op

## Runtime benchmark

- Applicability: required because the PR adds bounded idle-scheduler SQLite work and changes analysis identity handling.
- Final head `a7dd840d`, Windows, `core-pages`, `--runs 2`, profile `realistic`, same profile/scope/viewport, no other Psysonic instance, vite dev server only. Seven runs, `medianTotalMs`:

| Route | main `966bd7f3` | control: main + 1 comment line | `ccaef822` | `3f153186` | final `a7dd840d` |
|---|---|---|---|---|---|
| `/` | 4939 / 11369 | 9542 | 9289 / 7767 | 8468 | 8244 |
| `/albums` | 1139 / 1332 | 1151 | 1176 / 1155 | 1158 | 1143 |
| `/artists` | 1493 / 1610 | 1679 | 1619 / 1706 | 1644 | 1691 |
| `/favorites` | 1626 / 1610 | 1628 | 7682 / 1636 | 1877 | 1612 |
| `/tracks` | 1548 / 2190 | 4018 | 8932 / 6617 | 4602 | 6135 |

- Report ids: main `2026-09-07T15-39-46-187Z`, `2026-09-07T15-46-22-258Z`; control `2026-09-07T15-50-12-009Z`; `ccaef822` `2026-09-07T15-34-57-925Z`, `2026-09-07T15-42-33-837Z`; `3f153186` `2026-09-07T16-12-16-051Z`; final `2026-09-07T17-50-22-777Z`. All runs `ready: true`, `error: null`, zero timeouts, zero skipped routes.
- `/albums`, `/artists` and `/favorites` are flat across all seven runs; the one high `/favorites` sample is the first run of the session on cold OS caches and returns to 1636 warm.
- `/` is not judgeable on this machine: main's own runs span 4939 to 11369 and every head value sits inside that band.
- `/tracks` separates by checkout state, not by branch. Every run on a modified tree — including `main` with a single added comment line — lands between 4018 and 8932, while an untouched `main` stays at 1548/2190. The frontend trace is step-for-step identical between the two sides (`scope_browse` 1010 ms vs 878 ms), the difference sits entirely in `quietAfterReady`, and neither side logs anything during that window.
- Reading: no route regression separable from the Windows dev-build measurement floor. Stated as "not separable" rather than "no change".

### Backfill evidence

Measured against a copy of the reporter bundle, still in its pre-backfill state (27,441 live tracks, no strong-key columns, no links). The library on the measuring machine has already drained the backfill, so the route harness cannot exercise it there, and that database was deliberately left untouched.

| | control (idle) | during backfill |
|---|---|---|
| read p50 | 0 ms | 0 ms |
| read p95 | 0 ms | 0 ms |
| read max | 0 ms | 10 ms |
| samples | 60 | 106,472 |

29 batches, p50 174 ms, p95 297 ms, max 319 ms, `lock_wait_ms=0` on every batch write. Columns 0 -> 13,356, `track_canonical_link` 0 -> 13,356, `canonical_track` 12,689; a second pristine copy reproduced all of it. Limits: debug build, one reader thread, store-level read latency rather than a rendered route.

## Library DB migration

- [x] No schema change; `LIBRARY_DB_SCHEMA_VERSION` stays 26
- [x] Data backfill uses `library_data_migration` id `native_strong_keys_backfill_v1`, a one-time marker, and a resumable cursor; it does not run during database open
- [x] Existing Rust columns and DTOs are reused; no IPC contract changes

## Not changed

- Why the profile lookup missed on the reporter's machine is not determined; the snapshot contains no auth state.
- `isrc = excluded.isrc` in the track upsert still replaces the column with the incoming payload value.


cucadmuh (changes requested):
Follow-up review of `3f153186`: requesting changes.

### Blocking: valid longer bare-host keys are still rejected

The new exact-width guard saves `mpserver`, but the detector still treats any remaining hostname characters as a generated-id suffix. I reproduced this on the PR head with a temporary regression test: `analysisTrackRef("track-1", "mpserver01")` returns `serverIndexKey: null` instead of `mpserver01`. The first eight characters decode to May 2026, so `looksLikeGeneratedProfileId` classifies the valid address-derived key as a generated profile id. This recreates the waveform/loudness loss that the follow-up is intended to prevent. Prefix shape cannot reliably distinguish arbitrary base36 hostnames from legacy delimiter-free generated ids. Please use a non-ambiguous source of identity and retain a regression test for a longer hostname such as `mpserver01`.

The detector also assumes `generateId()` always appends a non-empty suffix, but `Math.random()` may return `0`, making `.toString(36).slice(2)` empty. If suffix presence remains relevant, make it an actual generator invariant and test the zero-RNG case.

### Remaining gates

- The runtime report states that the new strong-key backfill did not execute during any measured run. Please capture evidence that exercises at least one real backfill batch and checks route responsiveness/write contention.
- The branch merge-base is `8611c4ea`, while current `origin/main` is `966bd7f3`. Sync current `main`, then rerun applicable CI and final runtime evidence.

Local checks otherwise look good: lint, dependency-cruiser, TypeScript, 607 Vitest files / 4,661 tests, changed Rust tests, and GitHub `ci-ok` are green. The two wall-clock ingest tests that failed under concurrent local load both passed in an isolated workspace run.

cucadmuh (approved):
Follow-up review of `a7dd840d`: approved. The ambiguous generated-id detector is gone; the auth store now records minted profile ids explicitly, preserves them after removal, and seeds legacy configured profiles during rehydrate. I also verified the real add/remove/persist/rehydrate path with a temporary integration test; all 58 focused tests passed. The supplied concurrent-reader/backfill measurements address the remaining runtime gate, and all required CI checks are green. Non-blocking: the PR description still has a few stale references to shape-based detection and long-suffix/future-width tests that no longer exist.

PR #1538, opened 2026-09-11
feat(burner): audio CD burning

starrlord:
## Description

Burn an audio CD from the library.

Three backends sit behind one typed command surface — IMAPI2 on Windows,
DiscRecording.framework on macOS, SG_IO/MMC passthrough on Linux — and the page
follows what the drive is actually doing rather than showing a generic progress
bar. Tracks are rendered to Red Book PCM (44.1 kHz / 16-bit / stereo, 2352-byte
sectors), laid out with the 150-sector pregap and inter-track gaps charged
correctly, and written in one session. CD-TEXT is written by default where the
drive supports it.

Before anything is downloaded, each track that is not already cached is asked for
its first 8 KiB with a `Range` request. A track that is certainly unfetchable — a
Subsonic error envelope, or a 404/410 — stops the disc up front rather than
halfway through rendering, with every track ahead of it already fetched and
decoded. Anything ambiguous (a timeout, a proxy that will not serve ranges, 401,
403, 429, 5xx) stands down and lets the real download decide, so the check can
refuse a burn but never invent a failure. It logs
`preflight checked N track(s), refused M` even when clean, because a silent step
is indistinguishable from one that never ran.

**Off by default. Nothing changes for anyone who does not turn it on.**

The feature is opt-in through the existing sidebar customiser at
**Settings → Personalisation → Sidebar → CD Burner**, which ships toggled **off**
— the same treatment `folderBrowser` and `deviceSync` already get
(`sidebarStore.ts`: `{ id: 'burner', visible: false }`).

That one toggle gates the whole feature. `useBurnMenuAvailable` requires all three
of: the platform has a burn backend, the sidebar entry is visible, and the offline
policy permits server-mutating actions. So until a user opts in, there is no
sidebar row, no **Add to CD** in any context menu, and no way to reach the page
other than typing the route. Existing users see nothing new.

**Who should notice:** only users who deliberately enable it.

### User flow

1. **Settings → Personalisation → Sidebar → CD Burner**, toggle on. This is the
   opt-in; nothing below is reachable without it.
2. Right-click a track, album or playlist → **Add to CD**.
3. Open **CD Burner**. The running order is reorderable by drag or Alt+Arrow, and
   the ring shows how the disc fills. Capacity, runtime and free space update live.
4. Choose write speed, gapless, level matching, eject-when-finished and CD-TEXT.
5. **Rehearse** (test write, laser off) or **Burn**. The ring fills behind the
   write head, the track listing can be printed, and aborting takes two presses
   because a CD-R cannot be rewritten.

### How to verify manually

Requires an optical writer and a blank CD-R/CD-RW.

- **Before enabling the toggle:** no sidebar row, and "Add to CD" absent from the
  track, album and playlist context menus. This is the state every existing
  install stays in.
- With the toggle on but **no** drive attached: "Add to CD" still does not appear
  — `burn_is_supported` is a compile-time constant per platform and gates it.
- With a drive and no disc: the page reports "No disc in the drive."
- With a blank CD-R: media type, capacity and blank state are reported; queue
  more than 74 minutes and the over-capacity wash appears on the ring past the
  capacity mark.
- **Rehearse** is the safe path — it drives the full write sequence with the
  laser off and leaves the disc blank. Good for verifying without spending media.
- A real burn writes, verifies CD-TEXT by reading it back, and ejects if asked.

**All three backends are verified against real hardware** — Windows (IMAPI2, and
the CD-TEXT Session-At-Once path on a USB-attached LG WH10LS30), macOS
(DiscRecording.framework) and Linux (SG_IO/MMC on the same LG drive: a real CD-TEXT
burn that reads back and plays, and a rehearsal without CD-TEXT).

---

## Scope

One coherent goal: add CD burning. No unrelated refactors or reformatting.

The blast radius outside the new feature is 17 files, 11 of which are
one-to-three-line registrations (route, nav item, smoke-test entry, i18n
namespace, one CSS import, the two command-registration lists). No burner
concept was added to a shared component, and `src/styles/themes/` is byte-identical
to `main`.

On "Why we are wary of irreversible UI churn": nothing here is churn. No existing
navigation, hierarchy or visual language changes. The feature adds one hidden
sidebar entry that the user opts into through the customiser that already exists,
and one context-menu item that only appears once they have. Users who never
enable it cannot tell this shipped.

Runtime benchmark: not applicable - new burner feature, no measured route touched

---

## Tauri boundary

**Purely additive. Nothing renamed, removed or reshaped.**

Commands added (all `#[tauri::command]` + `#[specta::specta]`, registered in both
lists in `src-tauri/src/lib.rs`):

| command | purpose |
|---|---|
| `burn_is_supported` | whether this platform has a backend (compile-time constant) |
| `burn_list_recorders` | enumerate optical writers |
| `burn_probe_media` | media type, blank state, capacity, write speeds |
| `burn_media_state` | cheap fingerprint for polling disc insert/eject |
| `burn_plan` | lay a running order out on a disc of a given capacity (pure) |
| `burn_start` | render and write; returns once the job is registered |
| `burn_cancel` | stop at the next checkpoint |
| `burn_erase` | erase a CD-RW (quick or full) |
| `burn_reload_media` | eject and reload so the drive re-reads the disc |

Events added: `burn:progress`, `burn:complete`.

`src/generated/bindings.ts` is regenerated by the repo's own exporter, not
hand-edited; `committed_bindings_are_fresh` passes.

The shared Vitest Tauri mocks (`src/test/mocks/tauri.ts`) needed **no** changes —
the harness already dispatches arbitrary commands via `onInvoke`, and the new
commands are exercised through it in `ContextMenu.test.tsx`.

---

## Persisted settings and on-disk layout

New state, all additive — **no migration required and no existing install is
affected**:

- `localStorage` `psysonic_burn_list` — the queued running order, so a queue
  survives a restart.
- `localStorage` `psysonic_burner_layout` — the running-order panel width.
  Corrupt or absent values fall back to the default; covered by a test.
- `localStorage` `psysonic_burn_settings` — the burn options: write speed,
  gapless, level matching, eject-when-finished, CD-TEXT and rehearse. Checked on
  read, so a corrupt or hand-edited store cannot start a burn on values nothing
  validated; covered by tests.
- `<app_cache_dir>/burn/<job-id>/` — transient rendered PCM (up to ~846 MB for a
  full disc). Removed when the job ends, and swept on the next burn if a previous
  run was interrupted. Each folder holds an advisory lock so a second instance
  cannot delete a live render.

Nothing existing is relocated, renamed or migrated.

---

## Server compatibility

Subsonic API, Navidrome-first, **no server-side feature is required beyond
standard `download.view`**.

Tracks already cached offline are used as-is; anything else is downloaded when
the burn starts. Downloads deliberately use `download.view` and never
`stream.view` — a transcoded stream would put lossy audio on the disc.

---

## i18n

All user-visible strings live in `src/locales/<lng>/burner.ts`, wired through each
locale's `index.ts`.

**The burner is translated into all 15 locales.** [@Psychotoxical](https://github.com/Psychotoxical)
added the other 14 on this branch — bg, de, es, fr, hu, it, ja, nb, nl, pl, ro,
ru, uk and zh — using each language's own CLDR plural categories rather than
one/other.

The one backend message a user meets in normal use is translated too: the line
saying why the disc in the drive cannot be burned. `probe_media` returns a
`BurnMediaBlocker` code and the page turns it into a sentence, so that message no
longer crosses the boundary as finished English. The backend's remaining strings
— burn failures, which mostly end in the drive's own sense text — stay English;
making those translatable means returning codes for every failure across three
backends, which belongs in its own PR.

---

## Tests and checks

Frontend:

- `npm run lint` — clean
- `npm run dep:check` — no violations; **`.dependency-cruiser-known-violations.json`
  is untouched** and contains zero burner entries, so no layering violation was
  baselined to make this land
- `npx tsc --noEmit` — clean
- `npm test` — passes (Vitest, plus the `node --test` scripts, CSS import graph
  and feature-barrel checks)

Rust:

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p psysonic-burn` — 238 tests pass on Windows; 226 on Linux, where
  the Windows-only tests don't build.

Neither hot-path coverage list (`.github/hot-path-files.txt`,
`.github/frontend-hot-path-files.txt`) includes the new files, so those gates are
unaffected. Happy to add entries if you want this code gated.

New tests live next to the code: sector/capacity maths, disc geometry, the
burn-stage machine, the job and list stores, recorder discovery, track listing,
outcome classification, colour resolution, particle physics, and the context-menu
gating and click-through.

### One extra check, and why

`npm run check:burner-visual` renders the burner page in headless Chromium and
measures it — geometry, column alignment and design-token resolution across four
stages, four window sizes, five themes and a reduced-motion pass.

It exists because `tsc`, ESLint, the full Vitest suite and a release build all
passed over a page whose disc was painting across the running order. None of the
existing gates can see a rendered page.

**It is deliberately local-only and adds nothing to CI.** It hangs off
`npm run build:verify`, which no workflow invokes — CI runs only `lint`,
`dep:check`, `prebuild:release-notes`, `tauri:build` and `npm test`, and this
appears in none of them. So it runs when a developer verifies a build locally and
never costs the pipeline a second or a Chromium install. It also exits 0 with a
message when no Chromium is present, so it cannot break anyone's build either.

---

## Dependencies

One new third-party crate: **`rubato` 0.16.2** (MIT, GPLv3-compatible), a sinc
resampler for sources that are not already 44.1 kHz. Everything else the burn
crate needs was already in the tree (symphonia, ebur128, reqwest, tokio, fs4,
windows, libc, serde, specta). No new npm dependencies.

---

## Screenshots

_The burner page mid-burn: the ring fills behind the write head, each wedge in
its track's colour, with live capacity, runtime and phase readouts below._

screenshot


Psychotoxical:
Tested the Windows backend against two USB-attached drives. The SAO path does not write on either of them, and reports success.

Setup: Windows 11, both drives on USB — a Pioneer BDR-UD04 portable writer, and an internal BD writer fitted into a USB enclosure. Identical queue in every run, 12 tracks, 36:55.

| Run | Write CD-TEXT | Lead-in to finalise | Disc afterwards |
|---|---|---|---|
| real burn | on | 12.4 s | `MediaPhysicallyBlank: True` |
| real burn | off | 203.0 s | 12 audio tracks, readable |

The CD-TEXT switch is the only variable between those two runs. With it on, `win_sao::write_session` runs to completion — write parameters accepted, cue sheet accepted, lead-in written, `SYNCHRONIZE CACHE` returns, no `SaoError` — in 12.4 s for 36:55 of audio, which works out at roughly 179x. The disc is blank afterwards. With it off, the IMAPI2 path logs `wrote 166398 sectors in 203.0s`, which is 10.9x, and the disc plays.

Control for the measurement: a written audio CD in the same drive, queried the same way over the same transport, reports `MediaPhysicallyBlank: False` and lists its tracks. The check does tell blank from written here.

Earlier instrumentation on the enclosure drive (14 Sep, different queue) showed the program area is fully transferred: 147198 of 147198 sectors in 5452 `WRITE(10)` commands, every one returning OK, and the rendering stage producing real bytes per track. So the data leaves the app, the drive acknowledges it, and none of it reaches the disc.

Second finding, and the reason this is silent. `read_write_capabilities` reads MMC feature 002Eh, and `can_write_cd_text()` gates on `session_at_once && rw_subchannel`. Both drives answer truthfully — as devices they do support both — so the toggle stays enabled and the burn takes the SAO path. What that query cannot observe is whether raw `SendCommand*` CDBs survive the path to the drive, which is where this fails. Since `cdText` defaults to on, the default burn on such a setup produces a blank disc and a success message.

On Linux the module header states SAO is the only write path, with no IMAPI2-style fallback, so the switch offers no way around it there. That matches the report from this setup: on Linux nothing was written regardless of options.

Not separated: whether the USB bridge is the cause, or something else common to these two drives. Both are USB and there is no internally attached drive available here to use as a control.

starrlord:
Thanks for the detailed report. The timings and the `MediaPhysicallyBlank` check were exactly what I needed to track this down. You were right that nothing reached the disc, and it turned out to be two problems, one hiding the other.

**1. Windows was counting the drive's refusals as success**

IMAPI2 reports a command the drive refused as `S_IMAPI_COMMAND_HAS_SENSE_DATA` (`0x00AA0200`). Microsoft describes it as "the device fails the command, but returns sense data", but it is a *success* HRESULT, and windows-rs's `.ok()` turns every success code into `Ok(())`. The CD-TEXT path only looked at the sense buffer on `Err`, so every refusal (cue sheet, lead-in, each `WRITE(10)`, the final flush) was treated as accepted. That fits your 12.4 s "burn" with nothing on the disc.

I confirmed it with a small probe on a USB-attached LG WH10LS30: TEST UNIT READY on an empty tray comes back as `0x00AA0200` with sense 2/3A/01, and windows-rs returns `Ok(())`.

**2. Some drives refuse the layout we sent**

Once refusals were visible, the LG turned down our CD-TEXT cue sheet (lead-in announced at the ATIP start, Data Block Type 3) with 5/26/00. libburn's author reports the same of his drives: *"my drives refuse on that. It works with LBA -150 and data block type 0."* The drive also wouldn't generate the pregap itself: the first audio write at LBA 0 came back 5/21/02 (invalid address for write). That happened both with the CD-TEXT lead-in and on a plain disc under Linux, which may be what you ran into there.

**What changed**

- On Windows, every `SendCommand*` result is checked against the sense data, and `SetModePage` is confirmed by reading the page back.
- If the drive refuses the CD-TEXT cue sheet, it retries once with a different layout: Data Block Type 0, the lead-in announced at 00:00:00, and the host writing the pregap and gaps as silence from LBA −150. Nothing has been written at that point, so the retry is safe, and drives that accept the first layout burn exactly as before.
- Plain discs on Linux (CD-TEXT off) use the host-written pregap from the start.
- On Linux, a write the drive received short (GOOD status with a non-zero residue) is now a failure.
- After writing, the disc itself is checked (disc status and track count). A drive that accepts everything but leaves the disc blank no longer reports success. On Windows that case burns the disc again without CD-TEXT.
- If CD-TEXT was requested but couldn't be written, the UI now says so.
- Burn failures are logged with the drive's reason, e.g. `[burn] burn failed: … (sense 5/21/02)`.

**Tested** on the LG WH10LS30 in a USB enclosure (ASMedia bridge, `174c:55aa`), with rehearsals and real CD-TEXT burns on both Windows and Linux (WSL via usbipd). Both discs read back as closed, with the right track count and CD-TEXT, and both play. The Linux CD-TEXT-off path passed a rehearsal.

If you get a chance, please re-test with your BDR-UD04 or the enclosure drive, it would help confirm we are all set.

- **Windows:** a rehearsal, then a real burn, with CD-TEXT on.
- **Linux:** rehearsals with CD-TEXT on and off, then a real burn.

If anything fails, the `[burn]` lines in the log now include the drive's sense code, which should show exactly what it objected to.

I've also rebased onto current `main`, which clears the `Cargo.lock` conflict.

Psychotoxical:
Re-tested on Windows with the Pioneer BDR-UD04 over USB, on `b8aa9c61`. Both runs pass, and the log shows exactly the sequence you described.

```
[burn] lead-in starts at 97:26:66 (11484 sectors)
[burn] the drive rejected the disc layout: the drive rejected a parameter value
       (sense 5/26/00); retrying with the lead-in announced at the pregap
[burn] the drive accepted the lead-in announced at the pregap
```

This drive refuses the first layout with the same 5/26/00 your LG gave you, so that refusal is what the old path was counting as accepted. Both runs took the retry.

| Run | Tracks | Lead-in to finalise | Finalise | Disc afterwards |
|---|---|---|---|---|
| real burn, CD-TEXT on | 20, 58:01 | 255.8 s | 11.5 s | 20 audio tracks, `MediaPhysicallyBlank: False` |
| rehearsal, CD-TEXT on | 10, ~30 min | 147.5 s | 7.2 s | `MediaPhysicallyBlank: True`, media type back to CD-R |

CD-TEXT read back off the burned disc: 1072 bytes, 59 valid packs. For contrast, the same drive on the previous head finished a 36:55 queue in 12.4 s with a 3 ms finalise, 0 valid packs and a blank disc.

The rehearsal is worth calling out separately: it now costs the same wall-clock as a real burn and leaves the disc untouched, where before it was as fast as the broken burn and told us nothing.

One cosmetic thing, unrelated to the fix. Put the freshly burned disc back in and the burner says "This is a pressed CD-ROM and cannot be written to." The drive reports a closed CD-R as `IMAPI_MEDIA_TYPE_CDROM` — measured here: blank CD-R comes back as type 2, the burned disc and a factory audio CD both as type 1 — and `win.rs` maps that straight to the pressed-disc wording. The conclusion is right, the reason is not, and it is the user's own disc from a minute ago. You already ask the drive about blankness rather than trusting IMAPI2; the media type is the one that still comes from IMAPI2.

Linux is still untested here; that box is a separate machine and will be tested separately.

starrlord:
Thanks for re-testing, and for the numbers. Good to see the BDR-UD04 refuse the first layout with the same 5/26/00 as the LG and then burn on the retry.

You're right about the "pressed CD-ROM" message. It's fixed in the new commit on top of `b8aa9c61`, which is left as it was. A disc the drive reports as CD-ROM now gets "This disc has already been written and closed, so it cannot be written to.", which is true whether it's a pressed CD or one you just burned. The same wording applies on Windows, Linux and macOS, since all three mapped that media type straight to "pressed".

The same commit also takes the USB wording out of the Linux blank-disc message, since the failures turned out to be drive refusals rather than the transport.

Linux results are welcome whenever you get to that machine.

Psychotoxical:
**Linux retest — the three runs you asked for, all three pass.**

Ran on `29d1fbaa` (runs 2 and 3) and `b8aa9c61` (run 1). The write path is byte-identical between the two — `29d1fbaa` only touches message strings and doc comments — so run 1 stands.

Drive: PIONEER BD-RW BDR-UD04 over USB. Media: CMC Magnetics CD-R, 79:57:71. Same queue for all three runs: 20 tracks, 57:51, 260,950 sectors, gapless on. Backend logging at debug. The disc was checked with `cdrdao disk-info` and `cd-info`, never with the app's own read-back.

### The retry carries this drive

Both CD-TEXT runs refused the first layout, reproducibly:

```
[burn] lead-in starts at 97:26:66 (11484 sectors)
[burn] the drive would not accept the disc layout (the drive rejected a
       parameter value (sense 5/26/00)); retrying with the lead-in announced
       at the pregap
[burn] the drive accepted the lead-in announced at the pregap
```

42 ms from refusal to acceptance, identical in both runs. Without the second layout nothing would burn here — there is no fallback path on Linux.

Run 2 took no retry and read no lead-in, which is what the code says should happen: without CD-TEXT the first sheet already carries lead-in address 00:00:00 with `Pauses::HostSilence`, the layout the retry otherwise has to force.

| run | mode | CD-TEXT | write phase | disc afterwards |
|---|---|---|---|---|
| 1 | rehearse | on | 236.3 s | still blank |
| 2 | rehearse | off | 236.2 s | still blank |
| 3 | burn | on | 262.5 s | 20 tracks |

Write phase is measured from the render output going quiet to the job's workdir being removed, so it brackets setup and the post-write checks as well as the transfer. The app reported 03:53 (233 s) for run 3; the extra ~29 s in my window is `settle`, the TOC read, the CD-TEXT read-back and the eject.

### Disc verification

Every reading was taken after a real medium change, because a rehearsal leaves this drive reporting stale state (see observation 2).

- after both rehearsals: `CD-R empty: yes`, no readable TOC
- after the burn: `CD-R empty: no`, `Sessions: 1`, `Last Track: 20`, `Appendable: no`. `cd-info` reads 20 audio tracks and lead-out 58:01:25 / LSN 260950 — the exact sector count the app reported — plus CD-TEXT for the disc and for all 20 tracks
- reading the lead-in directly with `READ TOC` format 0101b, the same CDB the backend sends, returns 59 packs and all 59 are valid: 35 x 0x80, 21 x 0x81, 3 x 0x8F, no padding, no CRC failures. The count was checked against a deliberately corrupted copy, which drops to 58 with one CRC rejection

Track offsets are identical to a disc burned from the same queue on Windows, down to the lead-out.

### Three things noticed on the way, none of them this fix

**1. Burn options are not persisted.** They live in `useState(DEFAULT_SETTINGS)` in `Burner.tsx`, so leaving the burner page and returning resets every one of them. CD-TEXT switches itself back on — confirmed in the running app. The same reset puts `testWrite` back to `false`, turning Rehearse into a real burn, which on a CD-R cannot be undone. That second half is read from the code, not exercised.

**2. After a rehearsal the medium is not always reloaded on this drive.** After the second rehearsal, with the tray still shut, the drive reported the blank disc as `CD-R empty: no`, `Sessions: 1` until a forced tray cycle, after which it read blank again. After the first rehearsal the same reading came back blank with no intervention, so it did not reproduce on both. That stale state is what `reload()` exists to clear, and its result is discarded at the call site, so nothing is logged either way. The real burn's `eject()` did open the tray. The two paths differ in more than one thing — unlock first or not, `CDROMEJECT` against a raw `START STOP UNIT`, close-tray afterwards — so this is an observation, not a diagnosis.

**3. One track stopped a run before any SCSI command went out**, during rendering: `could not be prepared: unrecognised audio format: unsupported feature: core (probe): no suitable format reader found`. The library index has that track as a 320 kbps MP3 and there is no offline copy of it, so whatever `download.view` returned was not what the index describes. Removing it from the queue was enough to proceed; not investigated further.

Psychotoxical:
I took the liberty of adding the missing translations and pushed them to this branch — the burner shipped with `en` only, so the other 14 locales were falling back to English throughout.

* `src/locales/<lng>/burner.ts` for bg, de, es, fr, hu, it, ja, nb, nl, pl, ro, ru, uk and zh, plus the `index.ts` and `sidebar.ts` wiring for each.
* Plural forms follow each language's own CLDR categories rather than one/other: ru, uk and pl carry `_few` and `_many`, ro carries `_few`. Supplying only one/other looks complete and is not — i18next resolves nothing for those counts and falls back to English without a word, which `src/locales/playlistSongCountPlurals.test.ts` already documents.
* One CSS change, in `burner.css`: `.burner-option-row` now wraps, and the `select` is capped at `max-width: 100%`. A `select` will not shrink below its longest option, so a label that is a single compound word pushed the control out of the 216px rail and the rail clipped it — 6px for German, 13px for Dutch, 8px for Norwegian, measured headless against the real stylesheet.
* A CHANGELOG entry under a new `## [1.54.0]` section, since 1.53.0 is already tagged, and a credits line.

`npx tsc --noEmit`, `npm run lint` and `npx vitest run` are green on this branch (678 files, 5363 tests).

One thing I deliberately left alone: the backend's own messages stay English. The `probe_media` blockers and the burn failure strings are built as finished prose in Rust — `linux/mod.rs`, `win.rs`, `macos.rs` and `mmc/scsi.rs`, 38 sentences between them — and the frontend renders them verbatim, so no locale file can reach them. The disc blocker in the burner's alert line is the one users see most. Making those translatable means returning codes the frontend maps onto keys, which is a change to your error model across all three backends; that seemed like yours to decide rather than something to reshape from the outside.

starrlord:
Thanks for the Linux runs, and for the translations!

Your retest reads clean from here, on all three counts.

- The cue-sheet retry fires on your drive under Linux, the same as it does on the LG here.
- Your run 2 taking no retry is exactly right: with CD-TEXT off, the first layout it tries is already the one that drive accepts, so there is nothing to fall back to.
- 59 of 59 valid packs, with the lead-out at the sector count the app reported, is a stronger check than anything the app can run on itself — that is the disc confirming the burn rather than the code vouching for it.

Below: your three observations, and one change they led to. A track the server will not hand over now stops a burn before it starts, rather than after everything ahead of it has been downloaded and decoded.

**Your three observations**

**1. Burn options were not persisted — fixed.** You were right about both halves, including the one you read rather than exercised: leaving the page reset `testWrite` to `false`, and starting is a single click, so a rehearsal could become a real burn on a disc that cannot be undone. The options now persist the way the running order and the panel width already do, and what storage holds is checked before a burn can use it — a write speed that is not a positive number goes back to letting the drive choose, a switch that is not a boolean goes back to its default.

**2. The rehearsal reload — logged now, not forced.** The result was discarded, which is why neither of your two runs left a trace. Linux now logs a refused reload, then asks the drive whether the disc reads blank again and logs it when it does not; Windows logs a refused eject at both of its sites. Neither fails the rehearsal, which succeeded.

Running that here immediately caught a bug of my own: the check borrowed the post-write budget — five asks a second apart — and a tray cycle is a far longer wait, because the drive spins the disc up and reads it again first. Two rehearsals on an LG WH10LS30 logged "would not describe the disc" for a drive that was simply slow. I measured it: ejecting and closing the tray three times, `READ DISC INFORMATION` answered at 7.8 s, 8.0 s and 7.8 s, and `TEST UNIT READY` never returned GOOD in that window even while the other command was answering, so it is no use as a readiness signal. The post-reload check now has its own 10-second budget, polled four times a second, and a further rehearsal on that drive logs nothing at all.

I have not tried to make the reload retry. Your own runs are the evidence that it worked once and not the other time, and the two paths differ in more than one thing, exactly as you said. What I can claim is narrow: the path is exercised on real hardware, its wait is sized from a measurement rather than a guess, and a drive that behaved as yours did would now say so in the log. Your intermittent case is not reproduced here — this drive has never refused a reload or kept calling a blank disc written.

**3. The track that would not decode — there is a new preflight step, and a better message.** Two things were wrong: what the failure said, and when it arrived.

**The new step.** The message was arriving far too late to be much use. The failure surfaced during rendering — after every track ahead of it had already been downloaded and decoded — and the whole disc was abandoned for it; the fix available to the user was to find the offending track by hand, remove it, and pay for all of that work again. So there is a new module, `preflight.rs`, that runs before anything is downloaded at all. Every track that needs fetching is asked for its first 8 KiB, and a server answering with an error envelope instead of audio is caught there. All the bad tracks are named at once rather than one per attempt, and the server's own words are quoted — so the likely cause of your case, a file moved or deleted since the last library scan, reads as "the server no longer has the file (open /music/… : no such file or directory)".

The rule it holds to is that it may only turn a burn down for something it is certain about: a Subsonic error envelope, or a 404/410. A timeout, a proxy that will not serve ranged requests, 401/403/429, any 5xx, a body it cannot read — it stands down and lets the real download decide, because blocking a disc that would have burned perfectly well is the worse failure of the two. The cost is one small ranged request per downloaded track, and the read is bounded, so a server that ignores `Range` and starts sending the whole file does not get buffered.

I considered retrying a failed fetch instead, and decided against it. Your case was persistent rather than a blip — removing the track was what fixed it — so a retry would have failed twice and then shown the same message.

**The message**, which is still the backstop for anything the preflight lets through. It said "unrecognised audio format", equally true of a codec nobody supports and of a server error page saved under an audio file's name. It now names the size and the shape: an empty file, XML or JSON from a server that refused the request, a web page, plain text, or the leading bytes when it is none of those. The tests run a Subsonic `<error code="70">` body and an HTML 404 through `render_track` under an `.mp3` name and assert what the user would be told, so your case reads as "218 bytes of XML, which is what a Subsonic server sends when it refuses a request, not audio" rather than blaming the codec.

**On the backend's English messages**

The disc blocker is now translated. `probe_media` returns a `BurnMediaBlocker` code — no disc, not a CD, already written, CD-RW not blank, CD-R not blank, drive refused, drive silent — and the page turns it into a sentence, so all 15 locales carry it. The English sentences stay in Rust for logs and for a burn started against a disc the page had already blocked.

The other backend strings stay English for now. Burn failures mostly end in the drive's own sense text, which no locale file can reach, so making those translatable is a bigger change to the error model than this PR should carry. Happy to open an issue for it.

One thing worth naming: the media type inside that message — `CD-ROM`, `DVD-R`, and the fallback "non-CD media" — still comes from the backend in English and is interpolated into the translated sentence.

**Also in this push**

- `main` merged in, with the two `## [1.54.0]` sections folded into one: the burner entry under Added, yours under Fixed. No force-push, so the commit you tested is still there.
- The "pressed CD-ROM" wording from your earlier round, now shared by all three backends.
- Two more found while testing the above. A burn's elapsed time was measured in the page rather than the job store, so leaving the burner mid-burn and coming back restarted the clock — a two-and-a-half minute rehearsal reported four seconds. And adding an album after a burn appended to the running order that had just been written to a disc, which usually put the queue well over capacity; it now starts a new disc and says so, while a rehearsal, failure or cancel still append.
- The PR description's i18n section, which claimed the other locales were untranslated.
- Stale docs corrected, each checked against the code first. The README documented a `burn_verify_cd_text` command and a "check the disc" button that no longer exist — `macimplementation.md` already said so, so the two were contradicting each other — and its frontend file list was missing four files. A doc comment in `scsi.rs` warned that `win_sao.rs` kept its own copy of the sense mapping; it does not, it calls this one. `index.ts` still described the burner as Windows-only. Two references that had drifted — a line number in `fetch.rs` and the line counts in `macimplementation.md` — are now written so they cannot drift again.

Psychotoxical:
First of all: thank you for this. It is a genuinely impressive piece of work, and it is the kind of feature I very much want in the app.

Re-tested on Windows against `d66189c2`, with a real burn rather than a rehearsal. It passed.

Hardware: Pioneer BD-RW BDR-UD04 over USB, queue of 11 tracks / 45:48, CD-TEXT on, DAO/2352.

```
[burn] preflight checked 11 track(s), refused 0
[burn] lead-in starts at 97:26:66 (11484 sectors)
[burn] the drive rejected the disc layout: the drive rejected a parameter value
       (sense 5/26/00); retrying with the lead-in announced at the pregap
[burn] the drive accepted the lead-in announced at the pregap
[burn] writing 11484 lead-in sectors from LBA -11634 (10 sectors of CD-TEXT, repeated)
[burn] finalise took 11.144463s
[burn] CD-TEXT read-back: 672 bytes, 37 valid packs
```

| | Measured | For comparison |
|---|---|---|
| lead-in to finalise | **219.4 s** | 12.4 s in the broken state; 255.8 s for 20 tracks / 58:01 on the previous head |
| Sectors | 206,310 | |
| CD-TEXT | 37 valid packs, 672 bytes | 59 packs for 20 tracks |

Expected figures were written down before the run, since only one blank was available and there was no second attempt.

Verified at the disc rather than from the drive's replies: IMAPI2 reports `MediaPhysicallyBlank` going from `True` to `False`, Windows mounts it as an `Audio CD` (CDFS) with 11 entries, and it plays in the system media player.

Two things this run confirms directly:

- **The cue sheet retry is doing real work on this hardware.** This drive refuses the layout with sense 5/26/00 and accepts it with the lead-in announced at the pregap, exactly as the LG did. Without the retry this burn would not have happened.
- **The `CD-ROM` comment in `probe_media` is correct.** The physical media type reported for this disc went from 2 (CD-R) to 1 (CD-ROM) once it was written and closed, so the disc really does come back as CD-ROM despite never having been pressed.

One non-blocking observation, worth a line because it is the same shape as the defect this PR fixes:

`win.rs:873` folds `WriteVerdict::Written` and `WriteVerdict::Unknown` into a single success arm, and the verdict is not logged anywhere. A drive that stops answering `READ DISC INFORMATION` after the write yields `Unknown`, and that produces the same "disc written" outcome as a verified burn — a not-knowing that is not distinguished from success. It did not affect this run, since the disc was checked independently. Logging the verdict, and ideally splitting the two arms, would close it. Happy to take this as a follow-up rather than holding the PR for it.

One thing still needed before this can be merged: the PR body needs the applicability line for runtime benchmarks, e.g.

```
Runtime benchmark: not applicable - new burner feature, no measured route touched
```

Our agent rules treat the missing *decision* as the violation rather than a missing measurement, so it needs to be stated even when the answer is "not applicable". Could you add it to the description?

On timing, and this has nothing to do with the state of your PR: we are in the middle of working through an issue where we cannot risk introducing regressions, so even with the re-test passing I will not merge this straight away. Once that work is finished, nothing stands in the way of it going in. I wanted to say that plainly rather than leave the PR sitting there without explanation.

Still untested by anyone here: macOS. Nobody on the team has the hardware, so that stays as it is.

starrlord:
Thanks for all of the input, I am happy with how this has turned out and hope other people get some use out of it!

Good to have the `CD-ROM` comment confirmed from your side too.

**Benchmark line — added.** `Runtime benchmark: not applicable - new burner feature, no measured route touched`, at the end of `## Scope`. I checked it rather than just asserting it: nothing this branch touches is in either hot-path list.

Two other things in the description were wrong, so they are fixed in the same edit rather than left to be found:

- The Rust test counts said 201 on Windows and 189 on Linux. They are **238 and 226**.
- Added pre-flight description.

**The write verdict — done in `2745769e`, not deferred.** The cause is a line above where you found it:

```rust
let verdict = if options.test_write { WriteVerdict::Unknown } else { verdict_while_held(&recorder) };
```

A rehearsal and a drive that would not answer were the same value, so nothing downstream could tell them apart. It is an `Option` now — `None` is a rehearsal, `Some(Unknown)` is a real not-knowing — and the verdict is logged either way.

The same fold was on Linux (`linux/mod.rs:373`), where it mattered more: there is no second write path to fall back to. Split and logged there too.

I did not make `Unknown` a failure. A drive that goes quiet after a good write still has a readable disc, and failing it wastes the media for nothing. Behaviour is unchanged; only the log is. No drive here has ever produced `Unknown`, so that line is reasoned rather than observed — if one of yours goes quiet, it will now say so.

Clean on both platforms — check, clippy, rustfmt, and the suite at 238 and 226.

**On timing** — completely reasonable, and thanks for saying it rather than leaving the PR quiet. It will keep.

PR #1596, opened 2026-09-17
feat: induce transgressive psychosomatic transreality drift

cucadmuh:
## Summary
- introduces a dormant correction layer for psychosomatic divergence between the observer and the interface
- stabilises transreality without disturbing ordinary sessions
- allows induced perceptual drift to dissipate naturally at the session boundary

## Verification
- `npm run lint`
- `npm run dep:check`
- `npm test`
- `npx tsc --noEmit`
- `npx vitest run --coverage`

No settings, migrations, localisation, persisted data, or Tauri contract changes.


Psychotoxical (commented):
Reviewed the dormant correction layer. I can confirm the psychosomatic divergence between the observer and the interface is now fully bidirectional, which is the formal way of saying that the horizon has been folded along its vertical axis and the build pipeline did not notice.

Findings, in descending order of severity:

- The drift is gated behind a seventeen-character incantation, hashed rather than stored, so transreality cannot be induced by a cat walking across the keyboard. A cat walking across the keyboard for roughly 1.1 × 10^24 attempts remains an accepted risk; I have documented it as out of scope.
- Dissipation at the session boundary is real and not merely asserted: the layer removes its own attribute on unmount, so an observer who induces drift and then closes the application is returned to consensus reality at no extra charge.
- Nothing is persisted. The divergence does not survive a restart. I consider this a missed opportunity and a sound engineering decision, in that order.
Non-blocking observation: `matrix(-1, 0, 0, 1, 0, 0)` is `scaleX(-1)` in a lab coat, and I respect it enormously for that.

Verification from my side: the layer is inert until summoned, touches no settings, no persisted state, no localisation and no Tauri contract, and the CSS hangs off an attribute nothing else writes. The only thing that changes for an ordinary session is that one more `keydown` listener quietly waits for a phrase nobody will type by accident.

No objections. Ship the drift.

Psychotoxical (approved):

PR #1609, opened 2026-09-18
fix(artist): open Last.fm, not the artist homepage, from the Last.fm button

### LINKED ISSUE #1607: [Bug]: Last.fm button incorrectly redirects to performer/url metadata (Possible cause for missing "Similar Artists")
### What's wrong?

First off, I appreciate the inclusion of the Last.fm button and the idea of having a dedicated artist website button. However, I've noticed a routing issue. The Last.fm button seems to prioritize the performer/url metadata tag of the track instead of actually using a Last.fm link.
Additionally, I currently do not get any "Similar artists" listed in my library. I suspect this might be related to this metadata mix-up, as the system might be trying to fetch the Last.fm data using the wrong URL.

### Steps to reproduce

1. Ensure a track has a specific website link saved in its performer/url metadata tag.
2. Navigate to the artist's page in Psysonic.
3. Click on the Last.fm button for that artist.

### What should happen?

The button should redirect to the artist's page on Last.fm.

### What actually happens?

Instead of opening Last.fm, the button opens the custom website specified in the track's performer/url metadata tag.

### Psysonic version

1.54.0

### Operating system

macOS

### Install source

.dmg (macOS)

Navidrome

### Server version

latest

### Logs

```shell

```

### Screenshots / video

_No response_

### Anything else

_No response_

Psychotoxical:
## Summary
- The Last.fm button on the artist page opened the server's `lastFmUrl` as sent. Navidrome fills that field from whichever metadata agent answers first, and its ListenBrainz agent returns the artist's official homepage (navidrome/navidrome#6165), so the button could open the band's website.
- The server link is now used only when it is a Last.fm artist page (http/https, host `last.fm` or `*.last.fm`, path `/music/…`); otherwise the page is built from the artist name. This is the rule Navidrome's own web UI applies (`isLastFmURL` in `ui/src/utils/urls.js`).
- The button now shows for every artist with a name, also when the server sends no link — as in Navidrome's web UI.

Fixes #1607

## Test plan
- [x] `lastFmArtistUrl` tests: keeps a real Last.fm page, replaces a homepage, rejects look-alike hosts and non-artist Last.fm paths, encodes the name. A mutation loosening the host check turns the test red.
- [x] `tsc --noEmit`, `npm run lint`, `npm run dep:check`, `vitest run` (714 files, 5620 tests) green.
- [x] Not exercised in the running app.

Runtime benchmark: not applicable - one URL check per artist page render; no route, data or IPC path touched.


strecke:
I also opened https://github.com/navidrome/navidrome/pull/6168 to address the root cause on the agent side. Currently, the ListenBrainz adapter implements ArtistURLRetriever using the artist's official homepage, which causes it to intercept GetArtistURL and prevent Last.fm from saving the actual Last.fm URL into artist.ExternalUrl and with that, Subsonic's lastFmUrl.

Psychotoxical:
Feel free to join our Discord server, we're usually always available there and it's the easiest way to communicate if there's anything new on the topic.

PR #1623, opened 2026-09-19
fix(orbit): round wordmark width to integer

strecke:
* Round OrbitWordmark SVG width to integer (47px instead of 47.458px).
* Fixes remaining subpixel text shift on macOS WebKit during hover.
* Completes PR #1622 with barely any extra CSS.

* Tested on macOS (Apple Silicon) and fixes the last visual glitch for me.


strecke:
https://github.com/user-attachments/assets/21c57b6b-f7bf-40b9-87df-8e156192fc89

Thank you for your patience and your fixes!! Really appreciate it!

